PR #7, opened 2016-06-16
Support in-place template reinstalls - for testing

ttasket:
This doesn't yet prevent appvms from starting with invalid template during the reinstall, and doesn't deal with the Netvm setting problem.
For issue #2061

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -63,9 +63,11 @@ done
 
 # Prevent template upgrade - this would override user changes -
 # but do allow explicit template reinstalls
-if [ "$YUM_ACTION" == "reinstall" ] ; then
+if [ "$YUM_ACTION" == "reinstall" ] && [[ "$PKGS" == *"qubes-template-"* ]]; then
     TEMPLATE_EXCLUDE_OPTS=""
-else TEMPLATE_EXCLUDE_OPTS="--exclude=`rpm -qa --qf '%{NAME},' qubes-template-\*`"
+    echo "WARNING: Reinstalling a template will erase files in /home and /rw !"
+else
+    TEMPLATE_EXCLUDE_OPTS="--exclude=`rpm -qa --qf '%{NAME},' qubes-template-\*`"
 fi
 YUM_OPTS="$TEMPLATE_EXCLUDE_OPTS $YUM_OPTS"
 ALL_OPTS="$TEMPLATE_EXCLUDE_OPTS $ALL_OPTS"
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -68,12 +68,12 @@ if [ "$YUM_ACTION" == "reinstall" ] && [[ "$PKGS" == *"qubes-template-"* ]]; the
     echo "WARNING: Reinstalling a template will erase files in /home and /rw !"
     
     $ONEPKG=`cut -f 1 -d ' ' <<<$PKGS`
-    if [[ "$ONEPKG" == "qubes-template-"* ]] ; then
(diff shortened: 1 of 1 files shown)


marmarek:
Probably both issues should be handled in `template.spec`:
- netvm - save previous value (and whether it was default or not)
- AppVM starting - it should be easy not need this at all - by extracting `root.img` to some temporary name and then atomically renaming over original `root.img`

marmarek:
If you agree, I'd merge this PR and deal with those problems in the other one.

Hmm, on the other hand, changes to `template.spec` are slow to propagate - because templates are already in repository and contain %post script - no way to change those. And when user calls `reinstall` action, yum/dnf will try to get the same version, not newer. So maybe better to handle all of it here?

ttasket:
We can deal with it here, perhaps easily... Was going to look for a lock() mechanism against `qvm-start`, but can simply invalidate the template during that time by renaming root.img (and maybe *.conf) to something else as the first step. What do you think?

marmarek:
> but can simply invalidate the template during that time by renaming root.img to something else as the first step.

Yes, something like this. In fact, you already remove `root.img` in %post script of `template.spec` - basides the problem with old %post scripts in packages, it should be enough. So maybe indeed move it to `root.img.bak` here and if reinstallation succeed, remove it completely (and if not - restore it).

> (and maybe *.conf)

No need, it is autogenerated at every VM startup anyway.

ttasket:
Right, this is for the sake of non-refreshed %post scriptlets. Doing `rm -f` here might be better than a rename because any interruption won't leave the old root image behind. However, that would place deletion before the yum "Yes/No" prompt. :(

Other thing is, doing it here I'm assuming I'll need to extrapolate between the package name and what paths are in the qubes db.

marmarek:
> However, that would place deletion before the yum "Yes/No" prompt. :(

This is why I've proposed to rename it instead of removing. And also for cases like download fail etc. To not end up without any template...

> Other thing is, doing it here I'm assuming I'll need to extrapolate between the package name and what paths are in the qubes db.

`qvm-prefs fedora-23 root_img`

ttasket:
OK.

Is it realistic the user will try to specify *version* in the package name? Can I use '.' or '.noarch' as delimiter?

marmarek:
> Is it realistic the user will try to specify version in the package name? Can I use '.' or '.noarch' as delimiter?

As the template reinstall is very specific operation, I think you can expect to have exact template package name and nothing else. And fail for anything else matching `qubes-template-*` but you can't find existing template (like exact version, or wildcard).

Hmm, maybe better move this discussion back to qubesos/qubes-issues#2061 ?

ttasket:
I agree with 'fail on fancy' here... we can treat reinstall of templates and dom0 components as separate things.

So I'll require just a single matching package in case of qubes-template-*.

dom0-updates/qubes-dom0-update
```diff
@@ -173,8 +186,29 @@ elif [ -f /var/lib/qubes/updates/repodata/repomd.xml ]; then
...
+                    echo "OK"
+                else
+                    echo; echo "ERROR: Could not rename root.img"
+                    exit 1
+                fi
+            fi
+
+            yum $YUM_OPTS $YUM_ACTION ; RETCODE=$?
+            
+            if [[ -n "$BAK_TEMPLATE_ROOT" ]] ; then
```
marmarek: Shouldn't this check be about `$TEMPLATE_NETVM`?

ttasket: Intention is to repeat the check at ln191: It means, "are we handling a template?"

ttasket: See new PR https://github.com/ttasket/qubes-core-admin-linux/pull/1

PR #11, opened 2016-09-05
qubes-dom0-update: Show sync and download progress

rustybird:
Use `script` (part of util-linux) to fake a dumb terminal in the updatevm, so dnf will show sync and download progress indicators.


marmarek:
Good idea, thanks!

PR #17, opened 2016-12-13
Introduce $QREXEC_REMOTE_DOMAIN_*ID* in dom0, too

rustybird:
To be consistent with https://github.com/QubesOS/qubes-core-agent-linux/pull/31


marmarek:
See QubesOS/qubes-core-agent-linux#31 for reasoning.

PR #19, opened 2017-03-14
dracut: Do not fail if Xen components are built into the kernel

m-v-b:
Prior to this commit, if the Linux kernel's Xen-related components were
built into the kernel (as opposed to the use of kernel modules), then
the dracut module initialization would fail during the generation of the
initial ramdisk image.

This commit corrects this issue by using an if/then block.

Signed-off-by: M. Vefa Bicakci <m.v.b@runbox.com>


m-v-b:
I noticed that this commit was cherry-picked to your repository's master branch as https://github.com/marmarek/qubes-core-admin-linux/commit/e6e2404d24e2dc3fc84855a09bda91f6a652b7aa . For the record, this commit applies to the release3.2 branch as well, and it is helpful for users who compile their own kernels.

If there exists a possibility to cherry-pick this commit to the release3.2 branch, I would appreciate it. I will close this pull request for now though, to avoid cluttering our pull-request dashboards.

PR #22, opened 2017-03-31
Move qvm-xkill to different repo/pkg

jpouellet:
Tentatively qubes-desktop-linux-common.


jpouellet:
See https://github.com/QubesOS/qubes-issues/issues/2735

PR #25, opened 2017-04-27
.desktop parsing (receive appmenus) misc fixes

cfcs:
- Make `.desktop` match only `.desktop` and not `[wildcard character] desktop`

- Remove () from the list of accepted characters (this seemed to be causing trouble, do we need it?)

- Remove `%` from the whitelist of allowed characters in .desktop->Exec since the `%` character is interpreted as an escape by the [Desktop Entry specification](https://www.freedesktop.org/wiki/Specifications/desktop-entry-spec/), but not by the `pipes.quote()` function used to escape the string, which allows TemplateVMs to break out of the intended shell escaping.

- Change whitespace match for `ignore_rx` to have empty lines from the Desktop Entry files excluded.

Thanks to @DonnchaC for working with me on this.

### DIFF
--- a/appmenus-scripts/qubes-receive-appmenus
+++ b/appmenus-scripts/qubes-receive-appmenus
@@ -50,7 +50,7 @@ fields_regexp = {
     "GenericName": std_re,
     "Comment": std_re,
     "Categories": re.compile(r"^[a-zA-Z0-9/.;:'() -]*$"),
-    "Exec": re.compile(r"^[a-zA-Z0-9()_%&>/{}\"'\\:.= -]*$"),
+    "Exec": re.compile(r"^[a-zA-Z0-9()_&>/{}\"'\\:.= -]*$"),
     "Icon": re.compile(r"^[a-zA-Z0-9/_.-]*$"),
 }
 
@@ -142,8 +142,8 @@ def get_appmenus(vm):
 
     appmenus = {}
     line_rx = re.compile(
-        r"([a-zA-Z0-9.()_-]+.desktop):([a-zA-Z0-9-]+(?:\[[a-zA-Z@_]+\])?)=(.*)")
-    ignore_rx = re.compile(r".*([a-zA-Z0-9._-]+.desktop):(#.*|\s+)$")
+        r"([a-zA-Z0-9._-]+\.desktop):([a-zA-Z0-9-]+(?:\[[a-zA-Z@_]+\])?)=(.*)")
+    ignore_rx = re.compile(r".*([a-zA-Z0-9._-]+\.desktop):(#.*|\s*)$")
     for untrusted_line in untrusted_appmenulist:
         # Ignore blank lines and comments
         if len(untrusted_line) == 0 or ignore_rx.match(untrusted_line):


marmarek:
Thanks for the patch!
Take a look at https://www.qubes-os.org/doc/code-signing/ about signing your commits.

cfcs:
I don't have a computer with persistent storage and as such no good place to store a GPG key. :-(

I can generate a one-off key if that satisfies this requirement, or someone can copy my code and sign it with their key if they agree with the change?

marmarek:
Not even USB stick for a key (or yubikey/smartcard/other token)?

jpouellet:
> - Remove `%` from the whitelist of allowed characters in .desktop->Exec since the `%` character is interpreted as an escape by the [Desktop Entry specification](https://www.freedesktop.org/wiki/Specifications/desktop-entry-spec/), but not by the `pipes.quote()` function used to escape the string, which allows TemplateVMs to break out of the intended shell escaping.

I was concerned about this exact thing a while back and ended up auditing the propagation of Exec= lines as a result, and it did not appear to actually lead to code execution (at least in Xfce, and things which follow the desktop entry spec exactly and execve() themselves rather than passing to a shell - I did not audit KDE or awesome). I wanted to get around to actually addressing the problem properly before publishing my notes because they may have security implications, but regardless I am completely in favor of safer desktop-handling.

I believe an even more cautious approach is warranted, wherein no content from an untrusted domain ends up in Exec= at all, but rather is stored elsewhere and passed back as-is to the AppVM for which the desktop entry is being invoked without relying on any quoting to be safe.

cfcs:
I didn't manage to get it to execute code, and the extra parameters you can inject to the Qubes python script in dom0 (eventually) is after a "--", so it gets ignored. This seems a bit fragile, though, which I why I submitted this patch.

The `()` stuff made sections go away from the dom0 Xfce menu, and the `.`-as-wildcard is definitely a typo as well.

I do not make any claims that any of this has any real-world security implications, but I did spend time looking into that script, and these were my results. If you believe that the behaviour I described is indeed a design feature, that is cool with me - I have my own patchset that I run against Qubes anyway.
The patch adds three characters to two lines, and removes one character, and honestly you can take it or leave it, this is going to be my last input on this thread. I think @jpouellet makes a good point re: combinations of XDG menu parsing by other software (like KDE, awesome, i3, ...).

marmarek:
Removal of () will break things, for example:
```
whonix-ws/apps.templates/anondist-torbrowser_update.desktop:7:Name=%VMNAME%: Tor Browser Downloader (AnonDist)
whonix-ws/apps.templates/anondist-torbrowser_update.desktop:9:Comment=Download Tor Browser (AnonDist)
```
(and many windows application names)
Other changes are ok.

And yes, I agree with @jpouellet, `Exec=` line should be really be constructed by dom0 (in fact: GUI VM, when it will be separate). We plan to have a qrexec service that would take application id (desktop file base name in practice) as argument - with all the limitations about qrexec argument - and launch it. Internally it will use `qubes-desktop-run` command. This will also make easier to support `~/.local/share/applications` and such things. Hopefully we'll manage to have it in Qubes 4.0 (currently blocked on windows support).

PR #27, opened 2017-06-07
Fix root.img handling bug

tasket:
Per issue https://github.com/QubesOS/qubes-issues/issues/2848

Also moved "exit 1" block to end of conditional block for better chance of success.


marmarek:
Thanks @tasket!

PR #35, opened 2018-01-09
R3.2: PVH/HVM support

HW42:


marmarek:
Part of https://github.com/QubesOS/qubes-issues/issues/3460

PR #38, opened 2018-02-23
Make dir dom0-updates if not exists on UpdateVM

donob4n:
If for whatever reason UpdateVM misses "/var/lib/qubes/dom0-updates", dom0 updates fail.

https://github.com/QubesOS/qubes-issues/issues/3620

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -166,7 +166,7 @@ rm -f /var/lib/qubes/updates/errors
 echo "Using $UPDATEVM as UpdateVM to download updates for Dom0; this may take some time..." >&2
 
 # qvm-run by default auto-starts the VM if not running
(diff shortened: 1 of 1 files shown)


dom0-updates/qubes-dom0-update
```diff
@@ -166,6 +166,8 @@ rm -f /var/lib/qubes/updates/errors
 echo "Using $UPDATEVM as UpdateVM to download updates for Dom0; this may take some time..." >&2
 
 # qvm-run by default auto-starts the VM if not running
+qvm-run --nogui -q -u root  $UPDATEVM 'mkdir -m 775 /var/lib/qubes/dom0-updates/' || exit 1
```
marmarek: This on the other hand will fail if the directory exists. Add `-p`.

donob4n: Ok Marek, at least in Qubes 3.2 there was no problem since qvm-run ignores exit code without --pass-io, but it seems better this way.

PR #41, opened 2018-04-03
Srcrpm

fepitre:


fepitre:
@marmarek: Same remark than core-admin here and also for core-agent-linux.

PR #48, opened 2018-12-09
 Add qubes.repos.* qrexec services

strugee:
Originally filed at QubesOS/qubes-core-admin#246; initial review comments addressed there. (Thanks, @marmarek, for the helpful review!)

This is a prerequisite for QubesOS/qubes-issues#4550. I should note that while I introduced the files in the repo, I really don't understand how to introduce them into the build process, since there aren't already any other qrexec services in here...


strugee:
Ping @marmarek for review, whenever you get the chance.

strugee:
Oh, sounds super awesome! I'm not in a hurry, just thought it might have dropped off your radar since you've been pretty fast in the past. Which is much appreciated, by the way :)

strugee:
Hey, sorry for the long delay! I had final exams and I've been taking a break since then. (And then I forgot to push.)

Anyway, I added `ok` to the end of the script; if there's an error, it should crash the process before it reaches there. I also installed the services to `/usr/lib/qubes/qubes-rpc`; I had to create this directory. Hope that seems reasonable. I installed symlinks to them in `/etc/qubes-rpc` so they're picked up by qrexec. (At least, I think. I need to figure out how to point qubes-builder at this branch so I can actually make sure the RPM builds properly.)

> exception may reveal to much info

This is a good question. I guess I should just look up how to suppress exception messages in Python? That feels inelegant but also seems like the best way to prevent possible information leaks...

> I'd also think a little about the interface to be reusable for other distribution

This is a great point. I think `qubes.repos.List` should be okay; I'm not sure what the right thing to do for the repository name in APT would be though. I wonder if it would be reasonable to set that to just the empty string to indicate "not available".

For `qubes.repos.{Enable,Disable}` I think we'll just want to pass the full sources.list line as the parameter. Shouldn't be a problem to set that as the `id` in `qubes.repos.List` to match.

strugee:
@marmarek ping! I figured out how to use Qubes Builder so I was able to build the rpm. Verified the files were properly included with `rpm -q --filesbypkg -p`. Fixed the CI failures along the way.

strugee:
This shouldn't be merged as-is. In working on code that uses these services, I discovered they output an extra newline at the end. This is annoying for parsing.

strugee:
> This shouldn't be merged as-is. In working on code that uses these services, I discovered they output an extra newline at the end. This is annoying for parsing.

Fixed.

PR #51, opened 2019-10-06
qubes-dom0-update: send keyring

illikainen:
Send the dom0 keyring for QubesOS/qubes-core-agent-linux#187


marmarek:
Keyring is already part of the rpmdb sent there. No need to extract and re-import it.

PR #58, opened 2020-02-13
qubes-dom0-update: support 'list' and 'search' actions

pwmarcz:
The actions are already supported at the remote end
(qubes-download-dom0-updates.sh).

This is in order to implement Salt functions for dom0, see
QubesOS/qubes-issues#5252.

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -86,6 +86,11 @@ while [ $# -gt 0 ]; do
     shift
 done
 
+REMOTE_ONLY=
+if [ "CHECK_ONLY" == "1" ] || [ "$YUM_ACTION" == "list" ] || [ "$YUM_ACTION" == "search" ]; then
+    REMOTE_ONLY=1
+fi
+
 # Prevent implicit update of template - this would override user changes -
 # but do allow explicit template upgrade, downgrade, reinstall
 if [ "$YUM_ACTION" == "reinstall" ] || [ "$YUM_ACTION" == "upgrade" ] || [ "$YUM_ACTION" == "upgrade-to" ] \
@@ -203,7 +208,7 @@ tar c /var/lib/rpm /etc/yum.repos.d /etc/yum.conf /etc/dnf/dnf.conf 2>/dev/null 
 
 qvm-run $QVMRUN_OPTS --pass-io $UPDATEVM "script --quiet --return --command '/usr/lib/qubes/qubes-download-dom0-updates.sh --doit --nogui $(escape_args "${ALL_OPTS[@]}")' /dev/null" < /dev/null
 RETCODE=$?
-if [ "$CHECK_ONLY" == "1" ]; then
+if [ "$REMOTE_ONLY" == "1" ]; then
     exit $RETCODE
 elif [ "$RETCODE" -ne 0 ]; then
     exit $RETCODE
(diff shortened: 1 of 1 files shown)


dom0-updates/qubes-dom0-update
```diff
@@ -86,6 +86,11 @@ while [ $# -gt 0 ]; do
     shift
 done
 
+REMOTE_ONLY=
+if [ "CHECK_ONLY" == "1 " ] || [ "$YUM_ACTION" == "list" ] || [ "$YUM_ACTION" == "search" ]; then
```
marmarek: Trailing space in CHECK_ONLY value

dom0-updates/qubes-dom0-update
```diff
@@ -201,9 +206,17 @@ qvm-run --nogui -q $UPDATEVM 'rm -rf /var/lib/qubes/dom0-updates/etc' || exit 1
 tar c /var/lib/rpm /etc/yum.repos.d /etc/yum.conf /etc/dnf/dnf.conf 2>/dev/null | \
    qvm-run --nogui -q --pass-io "$UPDATEVM" 'LC_MESSAGES=C tar x -C /var/lib/qubes/dom0-updates 2>&1 | grep -v -E "s in the future"'
 
-qvm-run $QVMRUN_OPTS --pass-io $UPDATEVM "script --quiet --return --command '/usr/lib/qubes/qubes-download-dom0-updates.sh --doit --nogui $(escape_args "${ALL_OPTS[@]}")' /dev/null" < /dev/null
+CMD="/usr/lib/qubes/qubes-download-dom0-updates.sh --doit --nogui $(escape_args "${ALL_OPTS[@]}")"
+# Use 'script' to fake a terminal, so that we are shown sync and download
+# progress indicators. However, do it only if we are running in terminal
+# ourselves.
+if [ -t 1 ] && [ -t 2 ]; then
```
marmarek: Are you sure about checking stderr? This will make `qubes-dom0-update 2>/dev/null` behave interestingly ;)

dom0-updates/qubes-dom0-update
```diff
@@ -201,9 +206,17 @@ qvm-run --nogui -q $UPDATEVM 'rm -rf /var/lib/qubes/dom0-updates/etc' || exit 1
 tar c /var/lib/rpm /etc/yum.repos.d /etc/yum.conf /etc/dnf/dnf.conf 2>/dev/null | \
    qvm-run --nogui -q --pass-io "$UPDATEVM" 'LC_MESSAGES=C tar x -C /var/lib/qubes/dom0-updates 2>&1 | grep -v -E "s in the future"'
 
-qvm-run $QVMRUN_OPTS --pass-io $UPDATEVM "script --quiet --return --command '/usr/lib/qubes/qubes-download-dom0-updates.sh --doit --nogui $(escape_args "${ALL_OPTS[@]}")' /dev/null" < /dev/null
+CMD="/usr/lib/qubes/qubes-download-dom0-updates.sh --doit --nogui $(escape_args "${ALL_OPTS[@]}")"
```
marmarek: Just to be sure, have you verified the escaping? I think it is fine because `printf %q` (so `escape_args`) really escape special characters with `\`, not quote them, but worth double checking.
Most likely it will break for literal `'` inside argument (because you can't escape it with `\` within `'`), but I don't think we need to worry about it, it's very unlikely to be part of valid command line. And commands run in dom0 are trusted.

dom0-updates/qubes-dom0-update
pwmarcz: From what I understand, `qubes-dom0-update 2>/dev/null` will silence the errors, and also disable showing the progress bar. Am I right?

`script` will intercept both stdout and stderr, so if I just check stdout here, `qubes-dom0-update 2>/dev/null` will have no effect, which is also surprising.

My thinking was that as soon as someone starts doing something "custom" with the outputs, we should err on the side of caution, and disable our terminal hack.

marmarek: Ok, makes sense.

dom0-updates/qubes-dom0-update
pwmarcz: Yeah... from my experiments: spaces are passed correctly (they reach remote `dnf`), `"` also, and `'` break in remote bash.

If we will have more situations like this, I would replace `script` with some other way of ensuring terminal over qrexec. As a bonus, we could make `qvm-run bash` more comfortable. :)

dom0-updates/qubes-dom0-update
pwmarcz: Fixed.

PR #59, opened 2020-02-16
services: create /var/run/qubes-services/'SERVICE NAME' at boot

fepitre:


fepitre:
@marmarek: /etc/X11/xinit/xinitrc.d/qubes-sysinit.sh is not a good choice. Any better idea where to put it to be sure it starts before xdg/autostart and after qubesd?

PR #60, opened 2020-03-10
Fix kernel install script clobbering kernel params ending in default=

hal18000:
The original `sed` arguments caused the replacing of kernel parameters, for example, the default sleep mode parameter, `mem_sleep_default` would have it's value replaced with the kernel version being installed.

The solution is to only match the start of lines, as proposed by @marmarek 

Fixes QubesOS/qubes-issues#5716


hal18000:
Key is new and is likely waiting to propagate through keyservers. Key is available on GitHub as well as on [MIT's keyserver](http://pgp.mit.edu/pks/lookup?op=get&search=0x44086FD03B0775E1)  (published to SKS too, however it seems to be having difficulties right now)

PR #66, opened 2020-11-10
dom0-updates: double default max updates bytes

fepitre:
Add a safety mesure to ensure we have enought place on dom0 rootfs

### DIFF
--- a/dom0-updates/qfile-dom0-unpacker.c
+++ b/dom0-updates/qfile-dom0-unpacker.c
@@ -61,17 +61,6 @@ int main(int argc, char ** argv)
 		exit(1);
 	}
 
-	statvfs("/", &st);
-	// take a little margin with 90% of the free space
-	root_free_space = max(0, st.f_bfree * st.f_bsize * 0.90);
-
-	bytes_limit = min(root_free_space, DEFAULT_MAX_UPDATES_BYTES);
-
-	if ((var=getenv("UPDATES_MAX_BYTES")))
-		bytes_limit = atoll(var);
-	if ((var=getenv("UPDATES_MAX_FILES")))
-		files_limit = atoll(var);
-
 	uid = prepare_creds_return_uid(argv[1]);
 
 	incoming_dir = argv[2];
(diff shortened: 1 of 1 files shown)


dom0-updates/qfile-dom0-unpacker.c
```diff
@@ -49,13 +53,22 @@ int main(int argc, char ** argv)
...
 
+	statvfs("/", &st);
+	// take a little margin with 90% of the free space
+	root_free_space = max(0, st.f_bfree * st.f_frsize * 0.90);
+
+	bytes_limit = min(root_free_space, DEFAULT_MAX_UPDATES_BYTES);
+
 	if ((var=getenv("UPDATES_MAX_BYTES")))
-		bytes_limit = atoll(var);
+		bytes_limit = min(root_free_space, atoll(var));
```
marmarek, line 71: Explicit env variable should override the limit. This is specifically used during backup restore, where the backup is processed in parallel, so it will attempt to free the space in the meantime (and if we port https://github.com/QubesOS/qubes-core-agent-linux/pull/184/commits/7c261f45da20f7a57c18eff202d2d9dc07804e30 to dom0 version too, it's going to be guaranteed)

marmarek (changes requested):

dom0-updates/qfile-dom0-unpacker.c
```diff
@@ -49,11 +53,20 @@ int main(int argc, char ** argv)
...
 
+	struct statvfs st;
+	long long root_free_space;
+
 	if (argc < 3) {
 		fprintf(stderr, "Invalid parameters, usage: %s user dir [-v]\n", argv[0]);
 		exit(1);
 	}
 
+	statvfs("/", &st);
```
marmarek, line 64: One more thing: if /var/lib/qubes/updates (or whatever the target directory is) is on a separate partition, it should measure that. You can do that easily if you move setting the limit after `mkdir()` call below.

marmarek (changes requested):

dom0-updates/qfile-dom0-unpacker.c
fepitre: Done.

fepitre:
PipelineRetry

marmarek (approved):

PR #69, opened 2020-12-15
qfile-dom0-unpacker: support waiting for free space

marmarek:
Add support for -w option, which makes the extract process to wait for
enough free space in the target directory. This is especially useful
when restoring a backup, which is multi-stage asynchronous process -
wait for another process to process extracted files (and free the space)
before extracting more.

QubesOS/qubes-issues#5478

### DIFF
--- a/dom0-updates/qfile-dom0-unpacker.c
+++ b/dom0-updates/qfile-dom0-unpacker.c
@@ -10,6 +10,7 @@
 #include <unistd.h>
 #include <sys/fsuid.h>
 #include <errno.h>
+#include <limits.h>
 #include <libqubes-rpc-filecopy.h>
 
 #define DEFAULT_MAX_UPDATES_BYTES (4LL<<30)
@@ -49,6 +50,7 @@ int main(int argc, char ** argv)
 {
 	const char *incoming_dir;
 	int uid;
+    int i;
 	char *var;
 	long long files_limit = DEFAULT_MAX_UPDATES_FILES;
 	long long bytes_limit = DEFAULT_MAX_UPDATES_BYTES;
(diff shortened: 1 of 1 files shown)


dom0-updates/qfile-dom0-unpacker.c
```diff
@@ -49,6 +49,7 @@ int main(int argc, char ** argv)
 {
 	const char *incoming_dir;
 	int uid;
+    int i;
```
DemiMarie, line 52: Unintended tab?

dom0-updates/qfile-dom0-unpacker.c
```diff
@@ -90,7 +91,20 @@ int main(int argc, char ** argv)
...
 
 	set_size_limit(bytes_limit, files_limit);
-	if (argc > 3 && strcmp(argv[3],"-v")==0)
-		set_verbose(1);
+    for (i = 3; i < argc; i++) {
+        if (strcmp(argv[i], "-v") == 0) {
+            set_verbose(1);
+        } else if (strcmp(argv[i], "-w") == 0) {
+            if (i+1 < argc && argv[i+1][0] != '-') {
+                set_wait_for_space(atoi(argv[i+1]));
```
DemiMarie, line 99: `atoi` doesn’t detect errors, and has undefined behavior on overflow.  `strtoul` is better, but we might just want to check for `0` and `1` explicitly and reject other values.  This code also doesn’t handle `-w0` and `-w1`.

DemiMarie (changes requested):

dom0-updates/qfile-dom0-unpacker.c
marmarek: The other way around - the code is formatted with tabs, which I'm going to fix in separate PR.

dom0-updates/qfile-dom0-unpacker.c
marmarek: > reject other values

No, this option _does_ accept other values, specifically the free space that should be left (in bytes).

>  This code also doesn’t handle `-w0` and `-w1`.

Why should it? It isn't how _this program_ expects arguments. It is an internal tool not supposed to be called manually.

DemiMarie: Sorry, I didn’t think of that.  I still consider `strtoul` a better choice, since it has well-defined behavior and its return type (`unsigned long`) is what `set_wait_for_space` expects.

marmarek: Yes, `strtoul` is a good idea.

dom0-updates/qfile-dom0-unpacker.c
```diff
@@ -90,7 +92,28 @@ int main(int argc, char ** argv)
...
+    for (i = 3; i < argc; i++) {
+        if (strcmp(argv[i], "-v") == 0) {
+            set_verbose(1);
+        } else if (strcmp(argv[i], "-w") == 0) {
+            if (i+1 < argc && argv[i+1][0] != '-') {
+                unsigned long space_margin;
+
+                space_margin = strtoul(argv[i+1], &var, 0);
+                if (*argv[i+1] == '\0' || *var != '\0' ||
+                        (space_margin == ULONG_MAX && errno == ERANGE)) {
```
DemiMarie, line 104:
```suggestion
                errno = 0;
                space_margin = strtoul(argv[i+1], &var, 0);
                if (*var != '\0' || errno || *argv[i+1] < '0' || *argv[i+1] > '9') {
```

This ensures we don’t get a stale `errno` value, and checks for bogus leading whitespace.

DemiMarie (changes requested):
`strtoul` allows leading whitespace, which is not valid here.  Also, `errno` should be set to 0 before the call to `strtoul`.

PR #73, opened 2021-04-14
rpmcanon integration

DemiMarie:
So that next RPM vuln will not make us issue a QSB 🙂.

This is the long-awaited rpmcanon integration.  It requires the
qubes-rpm-oxide component to be enabled in the build, as that component
provides rpmcanon.  Therefore, this cannot be merged until that component has been signed by a trusted key.


marmarek:
PipelineRetry

marmarek (approved):

PR #79, opened 2021-08-28
Encrypt dom0 swap

DemiMarie:
This encrypts dom0 swap with a randomly generated key, which helps
prevent its contents from being recovered later.


DemiMarie:
Marking as draft because it is specific to the LVM install choice.

marmarek (changes requested):
I just found out I never clicked "submit review" here. Good thing that github kept the pending comment for a few years...

system-config/99-qubes-cryptsetup.conf
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Before=dev-mapper-%i.swap
+Requires=systemd-random-seed.service
+After=systemd-random-seed.service
+
+[Service]
+TimeoutSec=infinity
+ExecStartPost=/sbin/udevadm trigger /dev/mapper/%I
```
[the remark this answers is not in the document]
DemiMarie: Workaround for systemd bug that has since been fixed.

system-config/systemd-cryptsetup@swap.service
```diff
@@ -0,0 +1,35 @@
...
+# Do not stop this when trying to isolate a unit.
+IgnoreOnIsolate=true
+
+# Default dependencies for any encrypted volume.
+After=cryptsetup-pre.target systemd-udevd-kernel.socket systemd-random-seed.service
+
+# Ensure that this unit comes after its backing device,
+# and is shut down if its backing device is shut down.
+After=dev-qubes_dom0-swap.device
+BindsTo=dev-qubes_dom0-swap.device
```
marmarek, line 16: Even in LVM setup, the VG name can be different, and LV name can be different too...
Previous version also had this issue, but now it's all over the place instead of just StartExec line.

Maybe it's possible to rely on generators (and maybe the default ones will be enough)? Like, setup backing device via /etc/crypttab, and then use /dev/mapper/swap as the actual swap (via /etc/fstab).

marmarek (changes requested):

system-config/systemd-cryptsetup@swap.service
DemiMarie: That’s absolutely fine, and it’s what I do on my local system.  I have this line in `/etc/crypttab`:

```crypttab
swap /dev/qubes_dom0/swap /dev/urandom plain,swap,cipher=aes-xts-plain64
```

And this in `/etc/fstab` (modulo padding whitespace):

```fstab
/dev/mapper/swap none swap defaults 0 0
```

The reason that this PR does not take that approach is that these files are generated by Anaconda, which I have never touched before.  I can try, though.

DemiMarie (commented):
Is Anaconda the correct place to make this change?  I’m really unsure how to make changes to `/etc/fstab` and `/etc/crypttab`.

marmarek:
> Is Anaconda the correct place to make this change? I’m really unsure how to make changes to `/etc/fstab` and `/etc/crypttab`.

Yes. And the upgrade tool (https://github.com/QubesOS/qubes-issues/issues/9317). Sadly, Anaconda is quite complicated as storage goes, with logic split between anaconda itself and "blivet" package... But maybe it has encrypted swap support somewhere there already?
For convenience I just pushed my working branch for anaconda (upstream version + patches from our repo applied): https://github.com/marmarek/anaconda/tree/qubes-f41 (note - subject to force-pushes)
@fepitre do you have any hints? I think you touched storage in anaconda most recently...

PR #83, opened 2021-11-30
Fix dom0 updates if `repo_gpgcheck=1` is globally set

DemiMarie:
If `repo_gpgcheck=1` is set globally in dom0, updates break.  They will also break if DNF gets fixed to check signatures against the specific OpenPGP key specified in the repository file, instead of against the RPM keyring.  This fixes both problems.


marmarek (approved):
Approved as of 6261af4b7976ab5582187a026997d7b042ccf11b

PR #85, opened 2021-12-30
Try to make pciback more robust

DemiMarie:
Shell scripts are a terrible choice for critical code like this, but they are the best choice available, sadly.

### DIFF
--- a/dracut/modules.d/90qubes-pciback/module-setup.sh
+++ b/dracut/modules.d/90qubes-pciback/module-setup.sh
@@ -1,10 +1,20 @@
 #!/bin/bash
 
-install() {
+install () {
     inst_hook cmdline 02 "$moddir/qubes-pciback.sh"
     inst lspci
     inst grep
     inst awk
+    mkdir -p -m 0700 -- "$initdir/etc/usbguard"
+    mkdir -p -m 0755 -- "$systemdsystemunitdir/usbguard.service.d"
+    inst_multiple /etc/nsswitch.conf
+    inst_multiple /etc/usbguard/{qubes-usbguard.conf,rules.conf,rules.d,IPCAccessControl.d}
+    inst_multiple /etc/usbguard/rules.d/*
+    inst -l /usr/bin/usbguard
+    inst -l /usr/sbin/usbguard-daemon
+    inst /usr/lib/systemd/system/usbguard.service.d/30_qubes.conf
+    inst /usr/lib/systemd/system/usbguard.service
+    systemctl --quiet "--root=$initdir" -- enable usbguard.service
 }
 
 installkernel() {
--- a/dracut/modules.d/90qubes-pciback/qubes-pciback.sh
+++ b/dracut/modules.d/90qubes-pciback/qubes-pciback.sh
@@ -1,7 +1,7 @@
 #!/bin/bash --
 
 type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
-unset re HIDE_PCI usb_in_dom0 dev
+unset re HIDE_PCI usb_in_dom0 dev skip exposed
 
 usb_in_dom0=false
 
@@ -31,11 +31,21 @@ ws=$' \n'
 [[ $HIDE_PCI =~ ^[0-9a-f.:$ws]+$ ]] ||
     die 'Bogus PCI device list - fix your kernel command line!'
 modprobe xen-pciback 2>/dev/null || :
+dom0_usb=$(getarg rd.qubes.dom0_usb)
+case $dom0_usb in
+(*[!0-9a-f.:,]*) warn 'Bogus rd.qubes.dom0_usb option - fix your kernel command line'; dom0_usb=;;
+(*) dom0_usb=${dom0_usb//,/ } usb_in_dom0=true;;
+esac
 
 (
 set -e
 # ... and hide them so that Dom0 doesn't load drivers for them
 for dev in $HIDE_PCI; do
+    skip=false
+    for exposed in $dom0_usb; do
+        if [ "$dev" = "$exposed" ]; then skip=true; fi
+    done
+    if [ "$skip" = true ]; then continue; fi
     BDF=0000:$dev
     if [ -e "/sys/bus/pci/devices/$BDF/driver" ]; then
         echo -n "$BDF" > "/sys/bus/pci/devices/$BDF/driver/unbind"
--- a/dracut/modules.d/90qubes-pciback/qubes-pciback.sh
+++ b/dracut/modules.d/90qubes-pciback/qubes-pciback.sh
@@ -1,25 +1,49 @@
-#!/bin/sh
+#!/bin/bash --
 
 type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
+unset re HIDE_PCI usb_in_dom0 dev
 
-# Find all networking devices currenly installed...
-HIDE_PCI="`lspci -mm -n | grep '^[^ ]* "02'|awk '{print $1}'`"
+usb_in_dom0=false
 
-# ... and optionally all USB controllers...
 if getargbool 0 rd.qubes.hide_all_usb; then
-    HIDE_PCI="$HIDE_PCI `lspci -mm -n | grep '^[^ ]* "0c03'|awk '{print $1}'`"
+    # Select all networking and USB devices
+    re='0(2|c03)'
+elif ! getargbool 1 usbcore.authorized_default; then
+    # Select only networking devices, but enable USBguard
+    re='02'
+    usb_in_dom0=true
+else
+    re='02'
+    warn 'USB in dom0 is not restricted'
 fi
 
-HIDE_PCI="$HIDE_PCI `getarg rd.qubes.hide_pci | tr ',' ' '`"
+HIDE_PCI=$(set -o pipefail; { lspci -mm -n | awk "/^[^ ]* \"$re/ {print \$1}";}) ||
+    die 'Cannot obtain list of PCI devices to unbind'
 
+manual_pcidevs=$(getarg rd.qubes.hide_pci)
+case $manual_pcidevs in
+(*[!0-9a-f.:,]*) warn 'Bogus rd.qubes.hide_pci option - fix your kernel command line';;
+esac
+HIDE_PCI="$HIDE_PCI ${manual_pcidevs//,/ }"
+
+# XXX should this be fatal?
+ws=$' \n'
+[[ $HIDE_PCI =~ ^[0-9a-f.:$ws]+$ ]] ||
+    die 'Bogus PCI device list - fix your kernel command line!'
 modprobe xen-pciback 2>/dev/null || :
 
+(
+set -e
 # ... and hide them so that Dom0 doesn't load drivers for them
 for dev in $HIDE_PCI; do
     BDF=0000:$dev
-    if [ -e /sys/bus/pci/devices/$BDF/driver ]; then
-        echo -n $BDF > /sys/bus/pci/devices/$BDF/driver/unbind
+    if [ -e "/sys/bus/pci/devices/$BDF/driver" ]; then
+        echo -n "$BDF" > "/sys/bus/pci/devices/$BDF/driver/unbind"
     fi
-    echo -n $BDF > /sys/bus/pci/drivers/pciback/new_slot
-    echo -n $BDF > /sys/bus/pci/drivers/pciback/bind
+    echo -n "$BDF" > /sys/bus/pci/drivers/pciback/new_slot
+    echo -n "$BDF" > /sys/bus/pci/drivers/pciback/bind
 done
+) || die 'Cannot unbind PCI devices'
+if [ "$usb_in_dom0" = true ]; then
+    systemctl --no-block start usbguard.service
+fi
--- a/system-config/qubes-usb-rules.conf
+++ b/system-config/qubes-usb-rules.conf
@@ -1,0 +1,8 @@
+## Allow USB hubs.  Everything else requires this.
+allow with-interface match-all { 09:00:00 09:00:01 09:00:02 }
+
+## Allow HID devices, including keyboards and mice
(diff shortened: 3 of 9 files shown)


dracut/modules.d/90qubes-pciback/qubes-pciback.sh
```diff
@@ -1,25 +1,53 @@
...
-# Find all networking devices currenly installed...
-HIDE_PCI="`lspci -mm -n | grep '^[^ ]* "02'|awk '{print $1}'`"
+usb_auth_default=/sys/modules/usbcore/parameters/authorized_default
+usb_in_dom0=false
 
-# ... and optionally all USB controllers...
-if getargbool 0 rd.qubes.hide_all_usb; then
-    HIDE_PCI="$HIDE_PCI `lspci -mm -n | grep '^[^ ]* "0c03'|awk '{print $1}'`"
+modprobe usbcore authorized_default=0
+if getargbool 1 rd.qubes.hide_all_usb; then
```
marmarek, line 10: Please don't change the default, this PR isn't a place for that.

dracut/modules.d/90qubes-pciback/qubes-pciback.sh
```diff
@@ -1,25 +1,53 @@
...
+    # Select all networking and USB devices
+    re='0(2|c03)'
+elif grep -Fqe rd.qubes.hide_all_usb /proc/cmdline &&
```
marmarek, line 13: Use `getargbool` helper function.

dracut/modules.d/90qubes-pciback/qubes-pciback.sh
```diff
@@ -1,25 +1,53 @@
...
 fi
 
-HIDE_PCI="$HIDE_PCI `getarg rd.qubes.hide_pci | tr ',' ' '`"
+set -o pipefail
 
+HIDE_PCI=$(set -o pipefail; { lspci -mm -n | awk "/^[^ ]* \"$re/ {print \$1}";}) ||
+    die 'Cannot obtain list of PCI devices to unbind'
+
+manual_pcidevs=$(getarg rd.qubes.hide_pci)
+HIDE_PCI="$HIDE_PCI ${manual_pcidevs//,/ }"
```
marmarek, line 30: Can you add also an option to exclude specific device(s) (USB controller) from hiding? The use case would be to hide all the USB controllers except the one for HID, and use USBGuard to ensure it's HID-only. Maybe `rd.qubes.dom0_usb=` ?

marmarek (changes requested):

dracut/modules.d/90qubes-pciback/qubes-pciback.sh
DemiMarie: Done

system-config/usbguard.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+ConditionKernelCommandLine=|!rd.qubes.hide_all_usb
+ConditionKernelCommandLine=|rd.qubes.hide_all_usb=0
+ConditionKernelCommandLine=|rd.qubes.hide_all_usb=no
+ConditionKernelCommandLine=|rd.qubes.hide_all_usb=off
```
marmarek, line 5: and also:
`ConditionKernelCommandLine=|rd.qubes.dom0_usb`

marmarek (changes requested):

system-config/qubes-usb-rules.conf
```diff
@@ -0,0 +1,8 @@
+## Allow USB hubs.  Everything else requires this.
+allow with-interface match-all { 09:00:00 09:00:01 09:00:02 }
+
+## Allow HID devices, including keyboards and mice
+allow with-interface match-all { 03:00:00 03:00:01 03:00:02 03:01:01 03:01:02 }
```
3hhh, line 5: This implements https://github.com/QubesOS/qubes-issues/issues/6959 in its current state.

While I like seeing my issue addressed, I also fear that it'll break many current installs (incl. mine).

For example I have an external sound card which would be disabled, external USB HDs or smart cards with Qubes VMs would go away etc.

I could obviously fix all of that manually, but I guess many users would be surprised.

From my point of view this should be way more permissive by default (e.g. by allowing everything connected at boot time) and well-documented for users to create their own possibly more restrictive policy and/or there should be a big red warning in the changelog.

Btw I usually use 03:*:* for HID as a more permissive example. Also nice catch with `match-all`, which I hadn't seen before.

system-config/qubes-usbguard.conf
```diff
@@ -0,0 +1,36 @@
+RuleFile=/etc/usbguard/rules.conf
```
3hhh, line 1: I guess rpm still allows users to run their own usbguard config overwriting this one?

DemiMarie: This one is supplied by the distribution; at least on Fedora 35 the default is empty.

system-config/qubes-usb-rules.conf
marmarek: Good point @3hhh. I had a plan for external USB disks, but indeed there are more cases like this. So, I propose to make the whole thing conditional on `usbcore.authorised_default=0` in kernel cmdline - if not added explicitly, nothing will change. And the installer can add it when it considers it safe (for example not installing to the external USB disk), or the user can remove it back (if really needed...).

We will also include this in the upgrade instruction (to add the option if safe).

dracut/modules.d/90qubes-pciback/qubes-pciback.sh
```diff
@@ -1,25 +1,57 @@
...
+    # Select all networking and USB devices
+    re='0(2|c03)'
+elif ! getargbool 1 usbcore.authorized_default; then
+    # Select only networking devices, but enable USBguard
+    re='02'
+    systemctl start usbguard.service
+else
+    re='02'
+    warn 'USB in dom0 is not restricted'
 fi
```
marmarek, line 19: This needs to be a bit more robust. Specifically, enable usbguard in case of `rd.qubes.hide_all_usb usbcore.authorized_default=0 rd.qubes.dom0_usb=<some-specific-controller>`

dracut/modules.d/90qubes-pciback/qubes-pciback.sh
```diff
@@ -1,25 +1,57 @@
...
 
-# ... and optionally all USB controllers...
 if getargbool 0 rd.qubes.hide_all_usb; then
-    HIDE_PCI="$HIDE_PCI `lspci -mm -n | grep '^[^ ]* "0c03'|awk '{print $1}'`"
+    # Select all networking and USB devices
+    re='0(2|c03)'
+elif ! getargbool 1 usbcore.authorized_default; then
+    # Select only networking devices, but enable USBguard
+    re='02'
+    systemctl start usbguard.service
```
marmarek, line 15: In fact, you can simply set `usb_in_dom0=true` here instead of starting usbguard directly. Then, the below handling of `rd.qubes.dom0_usb` will set that too, so it will be started even if `rd.qubes.hide_all_usb` is set.

Ofc, then you need to restore starting usbguard if `usb_in_dom0=true`.

marmarek: Furthermore, this deadlocks for me. I guess it's some dependency cycle - this scripts is running rather early, for example before udevd starts.

Did it worked for you?

dracut/modules.d/90qubes-pciback/module-setup.sh
```diff
@@ -1,10 +1,19 @@
...
     inst awk
+    mkdir -p -m 0700 -- "$initdir/etc/usbguard"
+    mkdir -p -m 0755 -- "$systemdsystemunitdir/usbguard.service.d"
+    inst_multiple /etc/nsswitch.conf
+    inst_multiple /etc/usbguard/{qubes-usbguard.conf,rules.conf,rules.d,IPCAccessControl.d}
+    inst -l /usr/bin/usbguard
+    inst -l /usr/sbin/usbguard-daemon
+    inst /usr/lib/systemd/system/usbguard.service.d/30_qubes.conf
+    inst /usr/lib/systemd/system/usbguard.service
+    systemctl "--root=$initdir" -- enable usbguard.service
```
marmarek, line 16:
```suggestion
    systemctl --quiet "--root=$initdir" -- enable usbguard.service
```

marmarek:
It doesn't deadlock anymore, but also doesn't work. USB keyboard doesn't work neither at LUKS prompt, nor at lightdm prompt. It starts working only after I manually restart usbguard.

I was thinking usbguard is starting too late (after `sysinit.target`, which is after `local-fs.target` - so, too late for LUKS prompt), but:
a) it doesn't explain why USB keyboard doesn't work at lightdm
b) adding `DefaultDependencies=no` and `After=systemd-udevd.service` (and regenerating initramfs) did not helped

I've found messages like this:
```
Jan 12 17:28:19 dom0 usbguard-daemon[420]: USB Device Exception: SysFSDevice: (_sysfs_dirfd = open((G_sysfs_root + _sysfs_path).c_str(), O_PATH|O_DIRECTORY)) < 0: No such file or directory
Jan 12 17:28:19 dom0 usbguard-daemon[420]: SysFSDevice: (_sysfs_dirfd = open((G_sysfs_root + _sysfs_path).c_str(), O_PATH|O_DIRECTORY)) < 0: No such file or directory
Jan 12 17:28:19 dom0 usbguard-daemon[420]: USB Device Exception: SysFSDevice: (_sysfs_dirfd = open((G_sysfs_root + _sysfs_path).c_str(), O_PATH|O_DIRECTORY)) < 0: No such file or directory
Jan 12 17:28:19 dom0 usbguard-daemon[420]: SysFSDevice: (_sysfs_dirfd = open((G_sysfs_root + _sysfs_path).c_str(), O_PATH|O_DIRECTORY)) < 0: No such file or directory
```
which is weird, because I think sysfs is already mounted when udevd starts, but maybe I'm mistaken?

marmarek:
Please test the following scenarios:
 - no usb-related option set -> all USB controllers should remain in dom0, and all devices working there (for testing I simply check for `/dev/video0` - camera as a random non-HID device)
 - just `rd.qubes.hide_all_usb` set -> all USB controllers should be bound to pciback (`lsusb` in dom0 empty, USB keyboard _not_ working at LUKS prompt)
 - just `usbcore.authorized_default=0` -> all USB controllers in dom0, but only HID devices works (USB keyboard works at both LUKS prompt and later during normal system operations too)
 - `rd.qubes.hide_all_usb usbcore.authorized_default=0` - same as just `rd.qubes.hide_all_usb`
 - `rd.qubes.hide_all_usb usbcore.authorized_default=0 rd.qubes.dom0_usb=<BDF>` (where `<BDF>` is the only USB controller) - the USB controller remain in dom0, but only HID devices works (USB keyboard works at both LUKS prompt and later during normal system operations too)

For the purpose of this testing, detach USB controller from sys-usb.
For testing keyboard at LUKS prompt, replace keyfile path in `/etc/crypttab` with `none` (no need to remove the actual file) and rebuild initramfs. Make a backup of initramfs with the keyfile, just in case (you can switch to it in grub, or via the rescue system, if something goes wrong). You may need to press ESC twice to make plymouth show the passphrase prompt via KVM.

marmarek:
Also, please cleanup git history (squash some of those commits) and add a reference to https://github.com/QubesOS/qubes-issues/issues/2116 in relevant commits (especially about `rd.qubes.dom0_usb` option).

system-config/usbguard.service
```diff
@@ -0,0 +1,17 @@
...
+ConditionKernelCommandLine=usbcore.authorized_default=0
+ConditionKernelCommandLine=!usbcore.nousb
+Before=systemd-ask-password-plymouth.path systemd-ask-password-wall.path
+
+[Service]
+ExecStart=
+ExecStart=/usr/sbin/usbguard-daemon -f -s -c /etc/usbguard/qubes-usbguard.conf
+
+[Install]
+WantedBy=debug-shell.service rescue.target sysinit.target emergency.target
```
marmarek, line 17: `rescue.target` already pulls in `sysinit.target`, so no need to list the former. Also I believe `debug-shell.service` shouldn't be here (it's just enabled by relevant option on kernel cmdline, but normal targets are pulled in regardless).

marmarek (changes requested):
The service needs to be enabled (add to `75-qubes-dom0.preset` and use `systemctl preset usbguard.service` in `%post`). Otherwise it's stopped when transitioning out of initrd and not started later anymore.

marmarek:
After manually calling `systemctl enable usbguard.service`, it works for me in the main system. But still it doesn't at LUKS prompt...
According to logs, I see usbguard is started in initramfs early, before loading any usb controller drivers and even before udev. USB hubs are detected (`hub 2-0:1.0: 10 ports detected` etc), but not the actual devices. Interestingly, I don't see `kernel: usb X-Y: Device is not authorized for usage` message either - those are much later, in the main system already, and only for devices connected to those hubs.

dracut/modules.d/90qubes-pciback/module-setup.sh
```diff
@@ -1,10 +1,19 @@
...
-install() {
+install () {
     inst_hook cmdline 02 "$moddir/qubes-pciback.sh"
     inst lspci
     inst grep
     inst awk
+    mkdir -p -m 0700 -- "$initdir/etc/usbguard"
+    mkdir -p -m 0755 -- "$systemdsystemunitdir/usbguard.service.d"
+    inst_multiple /etc/nsswitch.conf
+    inst_multiple /etc/usbguard/{qubes-usbguard.conf,rules.conf,rules.d,IPCAccessControl.d}
```
marmarek, line 11: `inst_multiple` appears to _not_ copy directories recursively. Resulting `/etc/usbguard/rules.d` in initramfs is empty, so of course it can't work.
Adding separate `inst_multiple /etc/usbguard/rules.d/*.conf` made it work.

marmarek (approved):

PR #86, opened 2022-02-10
Make dom0 yum variables available in updatevm

Enteee:
In my dom0  I have repos installed which require custom yum variables.
This change is required to allow access to those variables on the update vm.

This works for me right now, but maybe there is a better way to solve this.


Enteee:
with all these failed tests, do I need to change something? or are the test just very unstable?

marmarek (approved):

Enteee:
cheers! :rocket:

PR #87, opened 2022-02-19
Make qubes.repo.* services handle new template repos location

marmarek:
Template repositories are now defined in /etc/qubes/repo-templates -
handle this location too.

QubesOS/qubes-issues#7252


tlaurion:
@marmarek in the goal of closing https://github.com/QubesOS/qubes-issues/issues/7252 this should be pushed directly to updates repo, no? Just had this issue migrating from q4.0 to q4.1 through qubes-dist-upgrade script.

PR #90, opened 2022-03-22
Better parsing of command-line options

DemiMarie:
This makes qubes-dom0-update aware of various command-line options that
take arguments, so it can process them accordingly.  It is not a perfect
solution; that would require using a C wrapper program to reject
abbreviated or unknown options.  It also uses bash regex matching
instead of the `echo | grep` anti-pattern, replaces an `==` with `=`,
adds a missing `--`, and unsets a variable that will be checked for
being set later.

The same approach used here could be used to refactor
qubes-gpg-client-wrapper if it were worth the time, which it probably
isn’t unless qubes-gpg-client-wrapper will see future use.

Marking as draft because I have not tested it yet.  I will test it once more time becomes available.


marmarek:
> This makes qubes-dom0-update aware of various command-line options that
> take arguments, so it can process them accordingly.

I don't like this part - the list can easily de-synchronize, there are also some actions that takes extra arguments (like repoquery). Having `--arg value` working for some but not the others is IMO worse than supporting only `--arg=value`.

> It also uses bash regex matching
> instead of the `echo | grep` anti-pattern, replaces an `==` with `=`,
> adds a missing `--`, and unsets a variable that will be checked for
> being set later.

This part is fine.

DemiMarie:
> > This makes qubes-dom0-update aware of various command-line options that
> > take arguments, so it can process them accordingly.
> 
> I don't like this part - the list can easily de-synchronize, there are also some actions that takes extra arguments (like repoquery). Having `--arg value` working for some but not the others is IMO worse than supporting only `--arg=value`.

What about explicitly detecting known bad cases and exiting with a an error message?  Ultimately I would like to do something like split-gpg1 and only allow command-line arguments we understand, so that we know we have not made any mistakes.  The DNF version in dom0 is frozen, so we should be able to do this unambiguously.  What do you think?

DemiMarie:
> > This makes qubes-dom0-update aware of various command-line options that
> > take arguments, so it can process them accordingly.
> 
> I don't like this part - the list can easily de-synchronize, there are also some actions that takes extra arguments (like repoquery). Having `--arg value` working for some but not the others is IMO worse than supporting only `--arg=value`.

Changed so that `--arg value` _never_ works (it results in an error).

marmarek:
> What about explicitly detecting known bad cases and exiting with a an error message?

Yes. this one is ok.

> The DNF version in dom0 is frozen, so we should be able to do this unambiguously.

There are several actions that are limited to the call in updatevm only (like search, or list), repoquery should also be one of them. For those, we don't have frozen set of supported options.

marmarek:
With this PR, `qubes-dom0-update` always fails. With rather unhelpful message: https://openqa.qubes-os.org/tests/37697#step/switch_pool/17

I tried to get more details by adding `--show-output --console`, but then got `--console xfsprogs must be written as --consolexfsprogs`.

So, please:
1. Fix handling `--console`
2. Fix error reporting on "Sending repository information to UpdateVM"
3. Fix the actual issue it has there.
4. Test it

And only then convert back to non-draft.

PR #91, opened 2022-04-04
Minor improvements to qubes-dom0-update

DemiMarie:
This uses bash regex matching instead of the `echo | grep` anti-pattern,
replaces an `==` with `=`, adds a missing `--`, and unsets a variable
that will be checked for being set later.  It also replaces some uses of
`[` with `[[`.

The same approach used here could be used to refactor
qubes-gpg-client-wrapper if it were worth the time, which it probably
isn’t unless qubes-gpg-client-wrapper will see future use.

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -365,8 +365,7 @@ fi
 
 if [[ "$REBOOT_REQUIRED" = 1 ]]; then
     echo "This upgrade requires system restart before doing anything else."
-    echo -n "Do you want to restart now? [Y/n] "
-    read answer
+    read -p 'Do you want to restart now? [Y/n] ' answer || :
     if [[ "$answer" != "n" ]] && [[ "$answer" != "N" ]]; then
         reboot
     fi
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -1,5 +1,6 @@
 #!/bin/bash --
 set -o pipefail
+unset YUM_ACTION UPDATEVM
 
 check_template_in_args () {
     local pkg
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -1,5 +1,5 @@
 #!/bin/bash --
-set -o pipefail
+set -euo pipefail
 unset YUM_ACTION UPDATEVM
 
 check_template_in_args () {
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash --
 
 find_regex_in_args() {
     local regex="${1}"
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -219,15 +219,15 @@ echo "Using $UPDATEVM as UpdateVM to download updates for Dom0; this may take so
 
 # qvm-run by default auto-starts the VM if not running
 readonly dom0_updates_dir=/var/lib/qubes/dom0-updates
-qvm-run --nogui -q -u root -- "$UPDATEVM" "mkdir -m 775 -p -- '$dom0_updates_dir'" || exit 1
-qvm-run --nogui -q -u root -- "$UPDATEVM" "chown -R -- user:user '$dom0_updates_dir'" || exit 1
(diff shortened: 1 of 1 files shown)


dom0-updates/qubes-dom0-update
```diff
@@ -1,13 +1,12 @@
...
-find_regex_in_args() {
-    local regex="${1}"
-    shift 1
-
-    for arg in "${@}"; do
-        if echo "${arg}" | grep -q -e "${regex}"; then
-            return 0
-        fi
+#!/bin/bash --
+set -euo pipefail
```
marmarek, line 2: Both `set -e` and `set -o pipefail` breaks this in several ways. At the very least, there are several places that operate on `$?` to handle various exit codes (not only errors) - with `set -e` all that will stop working. But then, there is `tar` call that is supposed to pack files on best-effort basis (not all those files may be present on every system) - pipefail breaks this.

marmarek (changes requested):
In it current shape, this PR makes `qubes-dom0-update` not functioning at all...

dom0-updates/qubes-dom0-update
```diff
@@ -222,75 +220,77 @@ echo "Using $UPDATEVM as UpdateVM to download updates for Dom0; this may take so
 
 # qvm-run by default auto-starts the VM if not running
 readonly dom0_updates_dir=/var/lib/qubes/dom0-updates
-qvm-run --nogui -q -u root -- "$UPDATEVM" "mkdir -m 775 -p -- '$dom0_updates_dir'" || exit 1
-qvm-run --nogui -q -u root -- "$UPDATEVM" "chown -R -- user:user '$dom0_updates_dir'" || exit 1
-qvm-run --nogui -q -- "$UPDATEVM" "rm -rf -- '$dom0_updates_dir/etc' '$dom0_updates_dir/var/lib/rpm'" || exit 1
-(cd / && exec tar -c -- var/lib/rpm etc/yum/vars etc/yum.repos.d etc/yum.conf etc/dnf/dnf.conf etc/pki/rpm-gpg/RPM-GPG-KEY-* 2>/dev/null) |
```
marmarek, line 228: You are reverting 37be9fbe610c67069ad4e59b595cc56a51da0514, I don't think it stopped being relevant (even if default install doesn't use it).

marmarek (changes requested):

dom0-updates/qubes-dom0-update
```diff
@@ -222,75 +220,81 @@ echo "Using $UPDATEVM as UpdateVM to download updates for Dom0; this may take so
...
-   qvm-run --nogui -q --pass-io -- "$UPDATEVM" "LC_MESSAGES=C tar -C '$dom0_updates_dir' -x &&
+qvm-run --no-gui -q -u root -- "$UPDATEVM" "mkdir -m 775 -p -- '$dom0_updates_dir'" || exit 1
+qvm-run --no-gui -q -u root -- "$UPDATEVM" "chown -R -- user:user '$dom0_updates_dir'" || exit 1
+qvm-run --no-gui -q -- "$UPDATEVM" "rm -rf -- '$dom0_updates_dir/etc' '$dom0_updates_dir/var/lib/rpm'" || exit 1
+dnf_conf=(var/lib/rpm etc/yum.repos.d etc/dnf/dnf.conf etc/pki/rpm-gpg/RPM-GPG-KEY-*)
```
marmarek, line 226: This now fails because `etc/pki/rpm-gpg/RPM-GPG-KEY-*` isn't getting expanded.

marmarek (changes requested):

dom0-updates/qubes-dom0-update
marmarek: Maybe include the whole dir?

marmarek:
I don't see specific error, but something is still broken - packages are not downloaded (or not found at all): https://openqa.qubes-os.org/tests/56662#step/switch_pool/15

DemiMarie:
> I don't see specific error, but something is still broken - packages are not downloaded (or not found at all): https://openqa.qubes-os.org/tests/56662#step/switch_pool/15

That is weird; I will deal with it later.

marmarek:
This already conflicts, and will conflict even more after merging https://github.com/QubesOS/qubes-core-admin-linux/pull/116. Closing for now. _If_ you are interested in working on it, please rebase (after #116 gets merged), make it actually work, and reopen.

PR #92, opened 2022-04-12
Fix qubes-dom0-updates.cron executable for cron

lubellier:
With fedora-32 the cron entries in `/etc/cron.daily` should get the executable bit, whereas cronie ignores these entries.

This PR is a partial solution for [#7437](https://github.com/QubesOS/qubes-issues/issues/7437).


marmarek:
> Would you mind changing this in the RPM spec file as well

It isn't necessary, RPM takes default from the actual file (unless overridden - which is not the case here).

marmarek (approved):

PR #96, opened 2022-04-16
Allow booting dom0 without Xen

DemiMarie:
When booting dom0 without Xen, the xen-pciback driver is (obviously) not
available.  Instead, use the vfio-pci driver.  This driver is normally
intended to reserve devices for use by userspace, but I believe it can
also be used to quarantine devices even if userspace will never use
them.


marmarek:
Superseded by #97

PR #97, opened 2022-04-16
dracut: do not fail native (non-Xen) boot

marmarek:
When booting Linux directly, binding devices to xen-pciback will fail.
Do not fail the boot in this case.

https://github.com/QubesOS/qubes-issues/issues/7390#issuecomment-1086745690

### DIFF
--- a/dracut/modules.d/90qubes-pciback/qubes-pciback.sh
+++ b/dracut/modules.d/90qubes-pciback/qubes-pciback.sh
@@ -56,7 +56,13 @@ for dev in $HIDE_PCI; do
     echo -n "$BDF" > /sys/bus/pci/drivers/pciback/new_slot
     echo -n "$BDF" > /sys/bus/pci/drivers/pciback/bind
 done
-) || die 'Cannot unbind PCI devices.'
+) || {
+    if [ -r /sys/hypervisor/type ] && [ "$(cat /sys/hypervisor/type)" = "xen" ]; then
+        die 'Cannot unbind PCI devices.'
+    else
+        warn 'Cannot unbind PCI devices - not running under Xen'
+    fi
+}
 if [ "$usb_in_dom0" = true ]; then
     info "Restricting USB in dom0 via usbguard."
     systemctl --quiet -- enable usbguard.service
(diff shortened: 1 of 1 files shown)


dracut/modules.d/90qubes-pciback/qubes-pciback.sh
```diff
@@ -56,7 +56,13 @@ for dev in $HIDE_PCI; do
     echo -n "$BDF" > /sys/bus/pci/drivers/pciback/new_slot
     echo -n "$BDF" > /sys/bus/pci/drivers/pciback/bind
 done
-) || die 'Cannot unbind PCI devices.'
+) || {
+    if [ -r /sys/hypervisor/type ] && [ "$(cat /sys/hypervisor/type)" = "xen" ]; then
```
DemiMarie, line 60: `cat` can be replaced by `read` here (see #96)

DemiMarie (changes requested):

dracut/modules.d/90qubes-pciback/qubes-pciback.sh
```diff
@@ -56,7 +56,17 @@ for dev in $HIDE_PCI; do
     echo -n "$BDF" > /sys/bus/pci/drivers/pciback/new_slot
     echo -n "$BDF" > /sys/bus/pci/drivers/pciback/bind
 done
-) || die 'Cannot unbind PCI devices.'
+) || {
+    hypervisor_type=
+    if [ -r /sys/hypervisor/type ]; then
+        read -r hypervisor_type < /sys/hypervisor/type
```
DemiMarie, line 62:
```suggestion
        read -r hypervisor_type < /sys/hypervisor/type || die 'Cannot determine hypervisor type'
```

marmarek: When you write such suggestion, you do realize this whole PR is about undoing your overly strict error checking?
(I can't think of a way why reading from existing /sys/hypervisor/type would fail, but also, can't miss the irony)

DemiMarie: :laughing: Yes, the irony is real

PR #111, opened 2022-12-14
prb/updater: more option

piotrbartman:
add option needed by gui

### DIFF
--- a/vmupdate/vmupdate.py
+++ b/vmupdate/vmupdate.py
@@ -144,7 +144,7 @@ def smart_targeting(app, args):
 def stale_update_info(vm, args):
     today = datetime.today()
     try:
-        last_update_str = vm.features.get(
+        last_update_str = vm.features.check_with_template(
             'last-updates-check',
             datetime.fromtimestamp(0).strftime('%Y-%m-%d %H:%M:%S')
         )
@@ -152,7 +152,7 @@ def stale_update_info(vm, args):
         if (today - last_update).days > args.update_if_stale:
             return True
     except qubesadmin.exc.QubesDaemonCommunicationError:
-        pass
+        return True
     return False
 
 
--- a/vmupdate/vmupdate.py
+++ b/vmupdate/vmupdate.py
@@ -144,6 +144,10 @@ def smart_targeting(app, args):
 def stale_update_info(vm, args):
     today = datetime.today()
     try:
+        if not ('qrexec' in vm.features.keys()
+                and vm.features.get('os', '') == 'Linux'):
+            return False
+
         last_update_str = vm.features.check_with_template(
             'last-updates-check',
             datetime.fromtimestamp(0).strftime('%Y-%m-%d %H:%M:%S')
@@ -152,7 +156,7 @@ def stale_update_info(vm, args):
         if (today - last_update).days > args.update_if_stale:
             return True
     except qubesadmin.exc.QubesDaemonCommunicationError:
-        return True
+        pass
     return False
 
 
--- a/vmupdate/vmupdate.py
+++ b/vmupdate/vmupdate.py
@@ -62,10 +62,13 @@ def parse_args(args):
     group.add_argument('--smart', action='store',
                        help='Default option. '
                             'Target all TemplateVMs with known updates or for '
-                            'which last update check was more than N days ago.'
-                            'Default value for N is 7 (one week).',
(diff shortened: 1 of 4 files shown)


vmupdate/vmupdate.py
```diff
@@ -90,17 +111,124 @@ def get_targets(args, app):
...
+                      if vm.klass in ('AppVM', 'DispVM')
+                      and vm.is_running()}
+        print("Following qubes CAN be restarted:",
+              ",".join((target.name for target in to_restart)))
+        return
+
+    # first shutdown templates to apply changes to the root volume
+    for template in templates_to_shutdown:
+        try:
+            template.shutdown()
```
marmarek, line 188: you need to wait for shutdown to complete before restarting appvms. Take a look at https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/events/utils.py#L42 for less busy-wait approach.

vmupdate/vmupdate.py
```diff
@@ -90,17 +111,124 @@ def get_targets(args, app):
...
+    for vm in to_restart:
+        if next(vm.connected_vms(), None) is None:
+            try:
+                vm.shutdown()
+                wait_for.append(vm)
+            except qubesadmin.exc.QubesVMError as exc:
+                print(exc, file=sys.stderr)
+        else:
+            try:
+                vm.kill()
```
marmarek, line 216: should be `vm.shutdown(force=True)` - it does graceful shutdown, but ignores dependencies check (like networking). In fact, you can always do `vm.shutdown(force=True)` (as you're going to restart it back) and don't need to check connected_vms()

vmupdate/vmupdate.py
```diff
@@ -90,17 +111,124 @@ def get_targets(args, app):
...
+                wait_for.append(vm)
+            except qubesadmin.exc.QubesVMError as exc:
+                print(exc, file=sys.stderr)
+
+    # wait for qubes to shut down
+    while any(vm.is_running() for vm in wait_for):
+        time.sleep(1)
```
marmarek, line 223: see the other comment about waiting for shutdown

vmupdate/vmupdate.py
```diff
@@ -50,14 +59,26 @@ def parse_args(args):
     group.add_argument('--all', action='store_true',
                        help='Target all non-disposable VMs (TemplateVMs and '
                             'AppVMs)')
-
+    group.add_argument('--n', action='store',
```
marmarek, line 62: `--n` looks weird, I liked `--smart` more, but maybe @marmarta has some better idea?

marmarek (changes requested):

vmupdate/vmupdate.py
piotrbartman: yes I also like `qubes-vm-update --smart` but `qubes-vm-update --smart 5` doesn't make sense to me. For now smart behavior is default so `qubes-vm-update` will do what we want, so first option is out. 
I'm open to proposition for a parameter name "how long without checking for updates is unacceptable"

marmarta: `--smart-time`, `--check-older`?

marmarta: Also chatGPT proposed '--update-if-stale N', and frankly it sounds like a good idea

marmarek (changes requested):

marmarek:
Most recent test run:
```
  File "/usr/lib64/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_update.py", line 626, in test_111_update_via_proxy_qubes_vm_update_cli
    self.update_via_proxy_qubes_vm_update_impl(
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_update.py", line 532, in update_via_proxy_qubes_vm_update_impl
    self.run_qubes_vm_update_and_assert(
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_update.py", line 471, in run_qubes_vm_update_and_assert
    self.fail(
AssertionError: qubes-vm-update return unexpected code: 1 in [0]
test-inst-vm1: ERROR (exception 'NoneType' object has no attribute 'put')
```

marmarek (approved):

PR #117, opened 2023-02-27
qubes-vm-update: capture SIGINT

piotrbartman:
Sending a SIGINT signal now causes a graceful shutdown, i.e. vms for which update has not started are canceled, but updates in progress are completed.

### DIFF
--- a/vmupdate/agent/source/status.py
+++ b/vmupdate/agent/source/status.py
@@ -22,8 +22,8 @@ from enum import Enum
 
 
 class Status(Enum):
-    WAIT = "wait"
     PENDING = "pending"
+    UPDATING = "updating"
     DONE = "done"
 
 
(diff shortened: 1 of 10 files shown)


vmupdate/agent/source/status.py
```diff
@@ -0,0 +1,53 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+# USA.
+from enum import Enum
+
+
+class Status(Enum):
+    WAIT = "wait"
+    PENDING = "pending"
```
marmarek, line 26: IMO this should be "in progress", or "updating" or similar. "pending" suggests it hasn't started yet.

vmupdate/qube_connection.py
```diff
@@ -137,19 +139,17 @@ def _copy_file_from_dom0(self, src, dest):
             with open(src, 'rb') as file:
                 proc = Popen(command, stdin=file)
```
marmarek, line 140: Not strictly related to this PR, but stdout/stderr should be redirected somewhere (either collected via subprocess.PIPE, or put to DEVNULL), otherwise it may disturb progress bar (at least color it red).
And also, why this one via `qvm-run` instead of python API, as the other calls?

marmarek (changes requested):

marmarek (approved):

PR #120, opened 2023-03-16
Misc improvements

DemiMarie:
See individual commit messages for details.

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -1,13 +1,9 @@
 #!/bin/bash
 
-find_regex_in_args() {
-    local regex="${1}"
-    shift 1
-
-    for arg in "${@}"; do
-        if echo "${arg}" | grep -q -e "${regex}"; then
-            return 0
-        fi
+has_template () {
+    local arg
+    for arg; do
+        case $arg in (qubes-template-*) return 0;; esac
     done
 
     return 1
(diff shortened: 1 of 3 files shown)


system-config/11-qubes-ignore-zvol-devices.rules
```diff
@@ -1,3 +1,4 @@
 # Determine if a zvol is a Qubes volume, then ignore it.
+KERNEL=="zd[0-9]*", SUBSYSTEM=="block", ACTION=="add|change", ENV{QUBES_IS_PART_OF_QVM_POOL}="1"
```
marmarek, line 2: This needs a comment. But also, it makes non-vm-related usage of zfs much more fragile. For example, this change on its own will break rootfs on zfs (if one manages to set it up), as `zvol_is_qubes_volume` script isn't present in initramfs. Such dependency needs to be documented somewhere.

dom0-updates/qubes-dom0-update
```diff
@@ -128,15 +124,18 @@ if [ "$CHECK_ONLY" == "1" ]; then
...
         exit 2
     fi
-    ONEPKG="$(echo "${PKGS[0]}" | sed -r 's/-[0-9]+(\.[0-9-]+)+(\.noarch)*$//')" # Remove version suffix
-    TEMPLATE=${ONEPKG#qubes-template-} # Remove prefix
+    if [[ "${PKGS[0]}" =~ ^qubes-template-(.*)(-[0-9]+)(\.[0-9-]+)+(\.[0-9a-zA-Z_]+)?$ ]]; then
+        TEMPLATE=${BASH_REMATCH[1]}
+    else
+        TEMPLATE=${PKGS[0]#qubes-template-}
+    fi
     echo "Redirecting to 'qvm-template ${YUM_ACTION=install} ${QVMTEMPLATE_OPTS[*]} $TEMPLATE'"
```
marmarek, line 137: If you change it below, change YUM_ACTION here too. It is set already when adding to PKGS.

marmarek (changes requested):

system-config/11-qubes-ignore-zvol-devices.rules
DemiMarie: > this change on its own will break rootfs on zfs (if one manages to set it up)

Root on ZFS will still work, as that does not use zvols.  Root and swap on a zvol should also work (though I have not tested it) due to:

```
# Tell udev that zvols are ready.
KERNEL=="zd[0-9]*", SUBSYSTEM=="block", ACTION=="add|change", ENV{SYSTEMD_READY}="1"
```

from `99z-qubes-mark-ready.rules`.

marmarek (approved):

PR #128, opened 2023-06-26
Skip double selecting of templates

piotrbartman:
fixes https://github.com/QubesOS/qubes-issues/issues/8162

### DIFF
--- a/vmupdate/vmupdate.py
+++ b/vmupdate/vmupdate.py
@@ -6,6 +6,7 @@ import asyncio
 import sys
 import argparse
 from datetime import datetime
+from typing import Set
 
 import qubesadmin
 import qubesadmin.exc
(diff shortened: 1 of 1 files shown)


vmupdate/vmupdate.py
```diff
@@ -87,24 +88,24 @@ def parse_args(args):
...
 
 
-def get_targets(args, app):
-    targets = []
+def get_targets(args, app) -> Set[qubesadmin.vm.QubesVM]:
+    targets = set()
     if args.templates:
-        targets += [vm for vm in app.domains.values()
-                    if vm.klass == 'TemplateVM']
+        targets.update(set([vm for vm in app.domains.values()
```
marmarek, line 94: Do you need inner `set()` when you add to set already?

vmupdate/vmupdate.py
```diff
@@ -114,17 +115,17 @@ def get_targets(args, app):
...
             )
     else:
-        targets += smart_targeting(app, args)
+        targets.update(smart_targeting(app, args))
 
     # remove skipped qubes and dom0 - not a target
     to_skip = args.skip.split(',')
-    targets = [vm for vm in targets
-               if vm.name != 'dom0' and vm.name not in to_skip]
+    targets = set([vm for vm in targets
```
marmarek, line 122: In fact, maybe earlier use of `set()` is not necessary at all, since you make it all set here?

marmarek (changes requested):

marmarek:
Oh, and please add "fixes" tag to the commit message too

marmarek (approved):

PR #129, opened 2023-06-26
add update related features to dom0

piotrbartman:
fixes https://github.com/QubesOS/qubes-issues/issues/8240

### DIFF
--- a/dom0-updates/qubes-dom0-updates.cron
+++ b/dom0-updates/qubes-dom0-updates.cron
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+qvm-features dom0 last-updates-check "$(date +'%Y-%m-%d %T')"
+
 if [ "$(qvm-features dom0 service.qubes-update-check || echo 1)" != 1 ]; then
     exit 0
 fi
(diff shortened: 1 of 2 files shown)


dom0-updates/qubes-dom0-updates.cron
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+qvm-features dom0 last-updates-check "$(date +'%Y-%m-%d %T')"
```
marmarek, line 3: Move this after the chunk below? Otherwise it will report updates were checked when in fact it was disabled.

marmarek (changes requested):

marmarek (approved):

dom0-updates/qubes-dom0-updates.cron
```diff
@@ -4,6 +4,8 @@ if [ "$(qvm-features dom0 service.qubes-update-check || echo 1)" != 1 ]; then
     exit 0
 fi
 
+qvm-features dom0 last-updates-check "$(date +'%Y-%m-%d %T')"
+
 # If no updates available - exit here
 qubes-dom0-update --check-only >/dev/null && exit
 RETCODE=$?
```
eaon, line 11: If I'm not mistaken then this sets the date even if the update check failed?

marmarek: I think you are correct. @piotrbartman can you look at it ?

PR #131, opened 2023-07-03
Add script to generate Xen unified kernel images

DemiMarie:
This is a prerequisite for secure boot.

Part of https://github.com/QubesOS/qubes-issues/issues/8206

### DIFF
--- a/uki-generate
+++ b/uki-generate
@@ -1,11 +1,12 @@
 #!/usr/bin/python3 -I
 import os
-import rpm
 import subprocess
 import sys
+import re
 from typing import List, Tuple
 
 def main(args):
+    section_re = re.compile(rb"\A *(?:0|[1-9][0-9]*) +([!-~]+) +([0-9a-f]{8}) +([0-9a-f]{16}) +[0-9a-f]{16} +[0-9a-f]{8} +2")
     alignment_mask = (1 << 21) - 1
     if len(args) != 6:
         print(f"Usage: uki-generate HYPERVISOR CONFIG KERNEL INITRAMFS OUTPUT", file=sys.stderr)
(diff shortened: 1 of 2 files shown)


uki-generate
```diff
@@ -0,0 +1,43 @@
...
+        f"--change-section-vma=.kernel={kernel_vma}",
+        f"--add-section=.ramdisk={initramfs}",
+        f"--change-section-vma=.ramdisk={initramfs_vma}",
+        "--",
+        hyp,
+        out,
+    ]
+    subprocess.check_call(cmdline, stdin=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
+if __name__ == '__main__':
+    main(sys.argv)
```
fepitre, line 43: Can you improve exception handling? Having full trackback may not be nice.

uki-generate
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/python3 -I
+import os
+import rpm
+import subprocess
+import sys
+from typing import List, Tuple
+
+def main(args):
+    alignment_mask = (1 << 21) - 1
+    if len(args) != 7 || args[1] != '--':
```
fepitre, line 10:
```suggestion
    if len(args) != 7 or args[1] != '--':
```

fepitre (changes requested):

uki-generate
```diff
@@ -0,0 +1,46 @@
...
+        f"--change-section-vma=.ramdisk={initramfs_vma}",
+        "--",
+        hyp,
+        out,
+    ]
+    subprocess.check_call(cmdline, stdin=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
+if __name__ == '__main__':
+    try:
+        main(sys.argv)
+    except subprocess.CalledProcessError as e:
```
fepitre, line 45:
```suggestion
    except (subprocess.CalledProcessError, ValueError) as e:
        print(str(e))
```

fepitre (changes requested):

uki-generate
DemiMarie: That should not happen.

DemiMarie: That should not happen.

fepitre: You raise ValueError in the code so it can happen and we need the error message.

DemiMarie: `ValueError` is thrown if one of the file sizes is nearly 2<sup>64</sup> bytes.  This is absolutely absurd and I don’t think we need to handle this situation.

marmarek: If something really shouldn't happen, make it an assert. Other exceptions should be properly handled.

marmarek:
@DemiMarie please add this script to packaging, makefile etc.

uki-generate
```diff
@@ -0,0 +1,48 @@
...
+import os
+import rpm
+import subprocess
+import sys
+from typing import List, Tuple
+
+def main(args):
+    alignment_mask = (1 << 21) - 1
+    if len(args) != 7 or args[1] != '--':
+        print(f"Usage: uki-generate -- HYPERVISOR CONFIG KERNEL INITRAMFS OUTPUT", file=sys.stderr)
```
marmarek, line 11: Please drop requirement of `--` (here, and generally in future tools you create). In this case, the tool doesn't even support any kind of options!

marmarek (changes requested):

uki-generate
DemiMarie: Ack.

uki-generate
```diff
@@ -0,0 +1,48 @@
...
+import subprocess
+import sys
+from typing import List, Tuple
+
+def main(args):
+    alignment_mask = (1 << 21) - 1
+    if len(args) != 6:
+        print(f"Usage: uki-generate HYPERVISOR CONFIG KERNEL INITRAMFS OUTPUT", file=sys.stderr)
+        sys.exit(1)
+    _, _, hyp, cfg, kern, initramfs, out = args
```
fepitre, line 13: Too much values to unpack.

fepitre (changes requested):

uki-generate
```diff
@@ -0,0 +1,69 @@
...
+        out = './' + out
+    output = subprocess.check_output([
+        "objdump",
+        "-WE", # do not use debuginfod
+        "--section-headers",
+        "--",
+        hyp,
+    ])
+    max_vma = 0
+    print(output.decode("UTF-8", 'surrogateescape'))
```
marmarek, line 27: This is leftover debug print, right?

DemiMarie: Yup!

PR #134, opened 2023-07-15
more logging

piotrbartman:
helpful to diagnose https://github.com/QubesOS/qubes-issues/issues/8163

### DIFF
--- a/vmupdate/qube_connection.py
+++ b/vmupdate/qube_connection.py
@@ -237,8 +237,8 @@ class QubeConnection:
     def _collect_stderr(self, proc) -> bytes:
         stderr = b""
         progress_finished = False
-        for untrusted_line in iter(proc.stderr.readline, ''):
-            if untrusted_line:
+        for untrusted_line in iter(proc.stderr.readline, None):
+            if untrusted_line is not None:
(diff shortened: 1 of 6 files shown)


vmupdate/qube_connection.py
```diff
@@ -248,21 +253,24 @@ def _collect_stderr(self, proc) -> bytes:
                         StatusInfo.updating(self.qube, progress))
                 else:
                     stderr += untrusted_line + b'\n'
-            else:
+            elif proc.poll() is not None:
```
marmarek, line 256: I can no longer reproduce the hang, I guess this fixed it :)

marmarek: I _guess_ the issue was stopping at the first empty line. If that's the case, it might be better to change the sentinel to `None` and this condition to `untrusted_line is None`. This way, you won't risk loosing last few lines of output printed just before process exit.

marmarek: I think both versions are wrong... What should really be here, is `b''` (not `''`) as the sentinel for the `iter()` call. I'll add a commit after a bit more testing.

marmarek (approved):

PR #139, opened 2024-01-05
Initial support for archlinux updates through pacman_cli

k4z4n0v4:


vmupdate/agent/source/pacman/pacman_cli.py
```diff
@@ -0,0 +1,68 @@
...
+            package, version = cols
+            packages.setdefault(package, []).append(version)
+
+        return packages
+
+    def get_action(self, remove_obsolete) -> List[str]:
+        """
+        Pacman will handle obsoletions itself
+        """
+        return ["-Syu"]
```
piotrbartman, line 68: If we cannot leave obsolete packages with pacman we should add info about that in [cli help](https://github.com/QubesOS/qubes-core-admin-linux/blob/e0b09228175c1246bf99e3fad76807b4e67abbb3/vmupdate/agent/source/args.py#L38) like `(ignored in Archlinux family)`

piotrbartman:
@k4z4n0v4 thanks for contribution! For me PR looks nice. @marmarek I think it should be merged after #141

vmupdate/agent/source/pacman/pacman_cli.py
k4z4n0v4: Sounds about right. If we exclude a few manual interventions that might be required per year or so, pacman should be able to handle the rest itself.

marmarek:
Generally looks good, but I'd like to get a successful update before merging. But that is currently impossible due to a missing pulseaudio-qubes update and that in turn didn't built due to hardware issue with the build system. Should be resolved next week.

k4z4n0v4:
I have made a version of pacman_api that works and correctly reports update/fetch/upgrade progress as much as possible, but It relies on proper reporting from pyalpm side.
Ill attach a modified PKGBUILD and a patch file for building pyalpm with the patches i made, if anyone wants to test.

PKGBUILD
```
# Maintainer : Rémy Oudompheng <remy@archlinux.org>

pkgname=pyalpm
pkgver=0.10.6.r9.g90941d0
pkgrel=1
pkgdesc="Libalpm bindings for Python 3"
arch=('i686' 'x86_64')
url="https://gitlab.archlinux.org/archlinux/pyalpm"
license=('GPL')
depends=('python>=3.2' 'pacman<=6.0.2')
source=("git+${url}.git")
md5sums=(SKIP)

pkgver() {
  cd "$pkgname"
  git describe --long --abbrev=7 | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
}

prepare(){
  cd ${srcdir}/${pkgname}
  patch -p1 < ../../patch1.patch
}

build() {
  cd ${srcdir}/${pkgname}
  python setup.py build
}

package() {
  cd ${srcdir}/${pkgname}
  python setup.py install --root=${pkgdir}
}
```
patch1.patch

```
diff --git a/src/options.c b/src/options.c
index 0c3d7aa..23cde25 100644
--- a/src/options.c
+++ b/src/options.c
@@ -339,8 +339,36 @@ void pyalpm_logcb(void *ctx, alpm_loglevel_t level, const char *fmt, va_list va_
   if (ret != -1) free(log);
 }
 
-void pyalpm_dlcb(void *ctx, const char *filename, off_t xfered, off_t total) {
+static int strendswith(const char *haystack, const char *needle)
+{
+	size_t hlen = strlen(haystack), nlen = strlen(needle);
+	return hlen >= nlen && strcmp(haystack + hlen - nlen, needle) == 0;
+}
+
+void pyalpm_dlcb(void *ctx, const char *filename, alpm_download_event_type_t event, void *data){
+	(void)ctx;
+
+	/* do not print signature files progress bar */
+	if(strendswith(filename, ".sig")) {
+		return;
+	}
+  alpm_loglevel_t level = ALPM_LOG_DEBUG;
+  PyObject *result;
+  char* log = "pyalpm_dlcb: unsupported download_event_type";
+	
+
+  if (event == ALPM_DOWNLOAD_PROGRESS){
+		dload_progress_event(filename, data);
+  } else {
+    result = PyObject_CallFunction(global_py_callbacks[CB_LOG], "is", level, log);
+    Py_CLEAR(result);
+	}
+}
+
+void dload_progress_event(const char *filename, alpm_download_event_progress_t *data){
   PyObject *result;
+	int xfered = data->downloaded;
+	int total = data->total;
   result = PyObject_CallFunction(global_py_callbacks[CB_DOWNLOAD], "sii", filename, xfered, total);
   if (!result) PyErr_Print();
   Py_CLEAR(result);
diff --git a/src/options.h b/src/options.h
index 911d3c9..97adc56 100644
--- a/src/options.h
+++ b/src/options.h
@@ -65,7 +65,8 @@ PyObject * option_remove_ignoregrp_alpm(PyObject *self, PyObject *args);
 
 /** Callback options */
 void pyalpm_logcb(void *ctx, alpm_loglevel_t level, const char *fmt, va_list va_args);
-void pyalpm_dlcb(void *ctx, const char *filename, off_t xfered, off_t total);
+void pyalpm_dlcb(void *ctx, const char *filename, alpm_download_event_type_t event, void *data);
+void dload_progress_event(const char *filename, alpm_download_event_progress_t *data);
 int pyalpm_fetchcb(void *ctx, const char *url, const char *localpath, int force);
 
 #endif
```

marmarek:
What about submitting the patch upstream? IIUC the proper place is at https://gitlab.archlinux.org/archlinux/pyalpm/-/merge_requests

k4z4n0v4:
As I said the upstream seems unmaintained. Plus the fix I have is not the proper, robust fix for the API change. It a lot more work, I just made it so that the part we interact with works.

marmarek:
> I'd like to get a successful update before merging

it worked :)

marmarek (dismissed):

k4z4n0v4:
Thanks @marmarek! Did you update with or without the progress bar?

marmarek:
For now I tested only the version without progress reporting, and merged only that.

k4z4n0v4:
I'll work on the code during my free time if I find anything to optimize / improve. What's the path forward you see regarding pacman_api, given the circumstances at hand with the pyalpm upstream?

vmupdate/agent/source/pacman/pacman_cli.py
```diff
@@ -0,0 +1,65 @@
...
+        self.package_manager: str = "pacman"
+
+    def refresh(self, hard_fail: bool) -> ProcessResult:
+        """
+        Use package manager to refresh available packages.
+
+        :param hard_fail: raise error if some repo is unavailable
+        :return: (exit_code, stdout, stderr)
+        """
+        cmd = [self.package_manager, "-Sy"]
```
christian-heusel, line 40: This is potentially problematic, as `pacman -Sy` without a related `-Su` or `-Syu` will produce a partial upgrade on the next `pacman -S <pkg>` as it is explained here (second point): https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported

Now if you can guarantee that there are no installations happening between `refresh()` and the update (`get_action()`?) that is not a problem, but if you cant you might want to consider something like `checkupdates` from `extra/pacman-contrib`, which does the sync against its own set of databases.

k4z4n0v4: Yeah I took that into account. 
Initially wanted to make pyalpm sync dbs to a temporary path of it's own to avoid that, but then in my testing there was a very short period between refresh and upgrade where the script releases the lock on the pacman DBs, which I think is highly unlikely to be enough for a package install. 

Either way to be extra safe I can theoretically move the refresh and the upgrade under one transaction, which will hold a lock on pacman DBs. A downside would be that it will make tracking the current progress harder. 
Or I can also just manually create and hold the  db.lock file until the very end of the script.

k4z4n0v4: Using checkupdates will also make it harder to track the update progress as we would have to wait for checkupdates to finish without being able to track progress, then do our own sync with progress tracking which would be an unintuitively long period for the end user without any (visible) progress.

k4z4n0v4: It somehow completely flew over my head that this is about pacman_cli, which has no progress reporting to begin with, sorry.

I'll experiment with checkupdates and artificially creating an holding a lock on the local DBs to see which works the best here. 

Will update with a patch.

k4z4n0v4: Technically speaking, I didn't really see the point of the `refresh()` step when it came to Archlinux in the first place, I just implemented it for the sake of interface completeness.  If @marmarek approves, I think the `refresh()` func for archlinux can be a no-op and it'll both solve this problem and reduce the running time of the whole upgrade.

marmarek: > I think the `refresh()` func for archlinux can be a no-op and it'll both solve this problem and reduce the running time of the whole upgrade.

I'm okay with that.

marmarek:
@christian-heusel could you help perhaps with the `pyalpm` issue discussed above? What is the best way to handle it?

k4z4n0v4:
Something broke pacman_cli, it now doesnt properly display the updated packages and instead just prints all packages 2 times.

christian-heusel:
> @christian-heusel could you help perhaps with the pyalpm issue discussed above? What is the best way to handle it?

Not an expert with pyalpm, sorry 😬 Maybe others can help, i.e. in `#archlinux-projects` 🤔

k4z4n0v4:
> Something broke pacman_cli, it now doesnt properly display the updated packages and instead just prints all packages 2 times.

Fixed with latest commit, pacman was calling `run_cmd` with `realtime=True` as it was the default

vmupdate/agent/source/pacman/pacman_cli.py
k4z4n0v4: Did just that with the latest commit.

marmarek:
> I have made a version of pacman_api that works and correctly reports update/fetch/upgrade progress as much as possible, but It relies on proper reporting from pyalpm side. Ill attach a modified PKGBUILD and a patch file for building pyalpm with the patches i made, if anyone wants to test.

Can you submit those modifications upstream? You said it is unmaintained, but I do see some movement in the repository, with most recent commit 2 weeks ago.

vmupdate/agent/source/pacman/pacman_api.py
```diff
@@ -0,0 +1,190 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+# USA.
+
+import os
+import pycman
+import pyalpm
+
+from source.common.stream_redirector import StreamRedirector
```
marmarek, line 26: StreamRedirector is gone, and not necessary anymore

k4z4n0v4:
I was initially happy to see some commits to pyalpm upstream a couple weeks ago, but they just turned out to be a modification to fix a test regression. The whole pyalpm lib is far behind the uptodate pacman/alpm library that now ships with archlinux. I can try submitting a pr for my minor changes upstream, but it is going to be a very brittle fix that's bound to fail eventually. I see  [at] jelly seems to be the most active contributor to pyalpm, Ill try to speak to them and see if they can give some help/guidance regarding all this.

k4z4n0v4:
Also can we close this pr for what it was originally? I think it'll be a mess if we start solving pacman_api issues here. I'll open a separate pr for the api, and we can just close this cause the cli works properly now I think.

k4z4n0v4:
The updater GUI currently doesn't work correctly with archlinux, the package still uses StreamRedirector. Can we merge this PR?

marmarek:
PipelineRetry

marmarek (approved):

PR #140, opened 2024-01-16
Work around a bug in PipeWire

DemiMarie:
PipeWire doesn't realize dom0 is a virtual machine and sets its quantums too low, resulting in xruns and unusably bad audio quality.


marmarek (changes requested):
packaged configs should be in /usr

DemiMarie:
Fixed

marmarek (approved):

PR #157, opened 2024-06-11
Simplification of updater CLI options

piotrbartman:
Refactoring CLI arguments to avoid unintuitive updater behavior along with suitable tests. 

Now, the selection of VMs for updating is based on two phases: preselection and selection. 
Preselection is based on the passed flags and selects VMs for further selection (e.g., all templates and some standalones). During selection, the VMs chosen for updating are those from preselection for which in case of flag:
`--update-id-available`: we know that updates are available; 
`--update-if-stale`, we know that updates are available or the VMs haven't been checked for updates in a long time; 
`--force-update`, all from the preselection.

TODO:
- [x] ci tests


marmarek (approved):

vmupdate/vmupdate.py
```diff
@@ -66,12 +66,18 @@ def main(args=None, app=qubesadmin.Qubes()):
...
+    ret_code_appvm, app_statuses = run_update(derived, args)
+    no_updates = all(stat == FinalStatus.NO_UPDATES for stat in app_statuses
+                     ) and no_updates
 
     ret_code_restart = apply_updates_to_appvm(args, independent, templ_statuses)
 
-    return max(ret_code_independent, ret_code_appvm, ret_code_restart)
+    ret_code = max(ret_code_independent, ret_code_appvm, ret_code_restart)
+    if ret_code == 0 and no_updates:
+        return 100
```
marmarek, line 79: This could use some documentation. I'll update integration tests as several of them failed on this one...
On a second thought, is it really a good idea to exit non-zero if everything is up to date already? I realize it may be useful information for some wrapping tools, so maybe have this behavior behind some option?
BTW, this applies both to the case where update was run and there was nothing to install, and to the case where update was skipped based on some flags (like no updates-available flag, and user didn't forced an update).

marmarek (approved):

PR #158, opened 2024-06-20
Use of software rendering by default

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/9268
For: https://github.com/QubesOS/qubes-gui-agent-linux/pull/214


marmarek:
Is it actually relevant for dom0? If you run GUI there, you have access to GPU in there.

marmarek:
We don't really support situation with GUI in dom0 but without graphics card there (like, GUI forwarded from dom0 to sys-gui-gpu). And even if we would, the primary need for graphics card in dom0 is window manager, which wouldn't be in dom0 in such case). So I don't think it's needed.

PR #159, opened 2024-06-24
qubes-dom0-update could check whether /boot is mounted

alimirjamali:
Fixes: https://github.com/QubesOS/qubes-issues/issues/885

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -29,6 +29,7 @@ if [ "$1" = "--help" ]; then
     echo "    --console    does nothing; ignored for backward compatibility"
     echo "    --show-output  does nothing; ignored for backward compatibility"
     echo "    --preserve-terminal  does nothing; ignored for backward compatibility"
+    echo "    --skip-boot-check  does not check if /boot & /boot/efi should be mounted"
     echo "    --switch-audio-server-to=(pulseaudio|pipewire) switch audio daemon to pipewire or pulseaudio"
     echo "                 it will be done after requested action (update by default)"
     echo "    <pkg list>   download (and install if run by root) new packages"
@@ -53,6 +54,7 @@ FORCE_XEN_UPGRADE=
 REBOOT_REQUIRED=
 DOWNLOADONLY=
 AUDIO_SWITCH=
+SKIP_BOOT_CHECK=
 # Filter out some dnf options and collect packages list
 while [ $# -gt 0 ]; do
     case "$1" in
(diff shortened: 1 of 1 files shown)


dom0-updates/qubes-dom0-update
```diff
@@ -234,6 +234,43 @@ if [ -n "$CLEAN" ]; then
...
+    local CHOICE
+    # No reason to check further if mount point is already mounted
+    awk -v PART="${1}" '{if ($2 == PART ) { exit 0 }} ENDFILE{exit -1}' < /proc/mounts
+    [[ ${?} -ne 0 ]] || return
+    # No reason to check further if mount point is not in fstab
+    awk -v PART="${1}" '!/^[ \t]*#/{ if ( $2 == PART ) { exit 0}} ENDFILE {exit -1}' < /etc/fstab
+    [[ ${?} -ne 0 ]] && return
+    # Ask user to manually mount partition if user is using GUI Updater
+    if [ ! -t 1 ]; then
+        echo "Could not decide about unmounted ${1} partition in non-interactive/GUI mode!"
```
marmarek, line 248: This doesn't allow overriding the check in non-interactive case. Maybe add some option for that?

dom0-updates/qubes-dom0-update
```diff
@@ -234,6 +234,43 @@ if [ -n "$CLEAN" ]; then
...
+                exit 1
+            fi
+            ;;
+        n|N) echo "Warning! Proceeding forward without mounting ${1}";;
+        a|A) echo Operation aborted!; exit 1;;
+        *) echo Invalid choice. Aborting!; exit 1;;
+    esac
+}
+
+if [ "$CHECK_ONLY" != "1" ]; then
```
marmarek, line 267: There are more cases when mounting is not necessary or desired: `REMOTE_ONLY=1`, `DOWNLOADONLY=1`. In fact `CHECK_ONLY` implies `REMOTE_ONLY` already.

marmarek (changes requested):

dom0-updates/qubes-dom0-update
alimirjamali: > This doesn't allow overriding the check in non-interactive case. Maybe add some option for that?

Is `--skip-boot` or `--skip-boot-check` good? And the help text:

```
--skip-boot       do no check if /boot & /boot/efi are mounted.
```

marmarek:
--skip-boot-check is okay.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

marmarek (approved):

PR #161, opened 2024-06-30
qubes-dom0-update: prevent concurrent execution

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/6345

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -187,6 +187,15 @@ if [ "$ID" != 0 ] && [ -z "$GUI" ] && [ -z "$CHECK_ONLY" ] ; then
     exit 1
 fi
 
+LOCKFILE="/var/run/qubes/qubes-dom0-update.lock"
+exec 9>"${LOCKFILE}"
+[ "$ID" == 0 ] && chgrp qubes "${LOCKFILE}" && chmod g+w "${LOCKFILE}"
+flock -n 9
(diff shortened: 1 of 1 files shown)


alimirjamali:
Moved the condition block down after checking if the script is run with root privileges (as it requires access to /var/lock)

marmarek (changes requested):
`--check-only` doesn't require root, and now it will fail. Maybe move the lock to /var/run/qubes (writable by the `qubes` group)? Or create a subdir in /var/lock (using systemd-tmpfiles) that is writable by the qubes group.

alimirjamali:
> Maybe move the lock to /var/run/qubes (writable by the `qubes` group)

This is a better idea. I even have used `/tmp/appname.lock` for non-root processes in the past. Amended PR will come in a second.

marmarek:
But there is still an issue - if the file was created by root, it won't be writable by the normal user. The /run/qubes dir has sgid bit set, so the group will be correct already, but the file won't be group-writable. Maybe adjust umask just for the time of opening the lock file? Or maybe open the lock for reading if already exists (to make it race-free, probably as a handling when opening for writing fails)?

alimirjamali:
I think it would be better to create the lock only if user is root. check-only should not worry about concurrent execution.

alimirjamali:
> Maybe adjust umask just for the time of opening the lock file? Or maybe open the lock for reading if already exists

I will implement both of the above approaches

marmarek (approved):

PR #162, opened 2024-07-27
Improve qubes-dom0-update messages

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/6126

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -137,8 +137,24 @@ case ${YUM_ACTION=upgrade} in
     ;;
 esac
 
+echo "Using $UPDATEVM as UpdateVM for Dom0" >&2
+
 if [ "$CHECK_ONLY" == "1" ]; then
     REMOTE_ONLY=1
+    echo "Checking for dom0 updates..." >&2
+else
+    case ${YUM_ACTION=upgrade} in
+    (upgrade|upgrade-to)
+        echo "Downloading updates. This may take a while..." >&2;;
+    (install|reinstall)
+        echo "Downloading packages. This may take a while..." >&2;;
+    (downgrade)
+        echo "Downgrading packages. This may take a while..." >&2;;
+    (list|search)
+        echo "Updating package lists. This may take a while..." >&2;;
+    (*)
+        echo "Performing ${YUM_ACTION}. This may take a while..." >&2;;
+    esac
 fi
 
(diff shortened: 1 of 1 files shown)


dom0-updates/qubes-dom0-update
```diff
@@ -137,8 +137,18 @@ case ${YUM_ACTION=upgrade} in
...
 esac
 
+echo "Using $UPDATEVM as UpdateVM for Dom0" >&2
+
 if [ "$CHECK_ONLY" == "1" ]; then
     REMOTE_ONLY=1
+    echo "Checking for dom0 updates..." >&2
+else
+    case ${YUM_ACTION=upgrade} in
+    (reinstall|upgrade|upgrade-to|downgrade)
```
marmarek, line 147: What about other actions like `install`? Maybe better rely on already existing `$REMOTE_ONLY` which I think already covers exactly the same difference?

marmarek (changes requested):

dom0-updates/qubes-dom0-update
alimirjamali: There is an special message for `list` & `search` further down in the PR. Since only package list is updated during list & search actions (and they are also included in `$REMOTE_ONLY`).

I broke the messages to more explanatory categories (and included `install`). I also added a `(*)` to catch all other actions (with appropriate message).

dom0-updates/qubes-dom0-update
```diff
@@ -137,8 +137,24 @@ case ${YUM_ACTION=upgrade} in
...
 if [ "$CHECK_ONLY" == "1" ]; then
     REMOTE_ONLY=1
+    echo "Checking for dom0 updates..." >&2
+else
+    case ${YUM_ACTION=upgrade} in
+    (upgrade|upgrade-to|downgrade)
+        echo "Downloading updates. This may take a while..." >&2;;
+    (install|reinstall)
+        echo "Downloading packages. This may take a while..." >&2;;
+    (downgrade)
```
marmarek, line 151: Now downgrade is twice...

marmarek (changes requested):

dom0-updates/qubes-dom0-update
alimirjamali: Ops. My bad.

It is fixed now.

marmarek (approved):

PR #164, opened 2024-09-04
vmupdater: fix #9072

piotrbartman:
add flags to apt conf to auto-resolve conffiles conflicts

fixes QubesOS/qubes-issues/issues/9072


vmupdate/agent/source/plugins/apt_keep_old_conffiles.py
```diff
@@ -0,0 +1,37 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+# USA.
+
+APT_CONF = "/etc/apt/apt.conf.d/01qubes-update"
```
marmarek, line 22: This will change also behavior for interactive calls later one, which isn't a good idea.
Maybe you can use `apt-get -o ..`, or `apt-get -c` (with some temp file)? Or equivalent via APT API?

marmarek (changes requested):

marmarek:
This has conflicts now.

marmarek:
@piotrbartman ping

marmarek (approved):

PR #166, opened 2024-09-07
dracut: include BDF of the device that failed to unbind

marmarek:


marmarek:
In fact, I'm not sure if it's going to work this way, as the loop executes in a subshell :(

PR #168, opened 2024-09-19
vmupdate: add support for nixos

evq:
this PR adds support for nixos updates based on the discussion in https://github.com/QubesOS/qubes-issues/issues/7992.

thanks for all the quick replies and advice @marmarek

#### notes:
- this uses a qubes specific wrapper script in order to allow for end user customization of the `nixos-rebuild` flags via the qubes update nix module. the possibility was mentioned that reading flags from a file might be preferred - however it seemed tidier this way and we can instead have the wrapper simply pass it's arguments on to `nixos-rebuild`.
- it also uses a qubes specific wrapper to retrieve currently installed packages. of the options available for listing installed package versions, `nix store diff-closures` seemed to be the best option however it is designed to diff. e.g. compare the previous system closure to the previous ones. the wrapper script first constructs an empty closure so that we can diff against it. again, it seemed tidier to abstract these details away.


evq:
will resign this with gpg later

vmupdate/agent/source/common/package_manager.py
```diff
@@ -276,7 +276,10 @@ def upgrade_internal(self, remove_obsolete: bool) -> ProcessResult:
         Just run upgrade via CLI.
         """
         cmd = [self.package_manager,
-               "--noconfirm" if self.package_manager == "pacman" else "-y",
+               *(
```
evq, line 279: i just extended the existing pattern but please let me know if you'd prefer I clean this up in some way. as-is package manager specifics are leaking into this common module.

marmarek: This is already fixed in https://github.com/QubesOS/qubes-core-admin-linux/pull/164, but it waits for resolving conflicts

vmupdate/agent/source/common/package_manager.py
```diff
@@ -227,17 +227,17 @@ def _print_changes(self, changes):
             for pkg in changes["updated"]:
                 result.out += self._print_to_string(
                     pkg,
-                    str(changes["updated"][pkg]["old"])[2:-2]
+                    ", ".join((changes["updated"][pkg]["old"]))
                     + " -> " +
-                    str(changes["updated"][pkg]["new"])[2:-2])
+                    ", ".join((changes["updated"][pkg]["new"]))
```
marmarek, line 232: This has a syntax error: not closed `(`

vmupdate/agent/source/common/package_manager.py
marmarek: FYI the other PR is merged now, you can rebase to resolve the conflict.

PR #172, opened 2024-11-09
Print & log argument errors instead of just logging

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/9565

### DIFF
--- a/vmupdate/vmupdate.py
+++ b/vmupdate/vmupdate.py
@@ -50,6 +50,7 @@ def main(args=None, app=qubesadmin.Qubes()):
     try:
         targets = get_targets(args, app)
     except ArgumentError as err:
+        print(str(err), file=sys.stderr)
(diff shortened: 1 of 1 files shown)


vmupdate/vmupdate.py
```diff
@@ -50,6 +50,7 @@ def main(args=None, app=qubesadmin.Qubes()):
     try:
         targets = get_targets(args, app)
     except ArgumentError as err:
+        print(str(err))
```
marmarek, line 53: print to stderr?

marmarek (changes requested):

vmupdate/vmupdate.py
alimirjamali: Done

marmarek (approved):

PR #184, opened 2025-02-28
Update firmware packages from bookworm-backports on Debian 12

marmarek:
Firmware packages in standard Debian repo are too old for recent
kernels, and recent hardware in general. This affects all kernel-latest
versions, but also soon-to-be-default kernel-6.12. This is at least
necessary to support wifi network in NovaCustom laptops, but other
modern systems need that too.
Fresh template builds have firmware from bookworm-backports already
selected, but do that for older versions too.

Thanks @aronowski for the hint on apt properties.
Fixes QubesOS/qubes-issues#9815

### DIFF
--- a/vmupdate/agent/source/plugins/bookworm_backports.py
+++ b/vmupdate/agent/source/plugins/bookworm_backports.py
@@ -37,6 +37,13 @@ Pin-Priority: 600
 
 """
 
+prefs_pipewire_data = """\
+Package: src:pipewire
+Pin: release n=bookworm-backports
+Pin-Priority: 600
+
+"""
+
 
 def add_backports_repo():
     # find URL flavor used for deb.debian.org
@@ -82,16 +89,19 @@ def check_package_not_from_backports(package):
 
 def bookworm_backports(os_data, log, **kwargs):
     """
-    Update firmware packages from backports repository.
+    Update firmware and/or pipewire packages from backports repository.
 
     https://github.com/QubesOS/qubes-issues/issues/9815
+    https://github.com/QubesOS/qubes-issues/issues/8560
+    https://github.com/QubesOS/qubes-issues/issues/8916
     """
     if os_data.get("codename", "") == "bookworm":
         # check what packages need to be updated to backports version
         update_firmware = check_package_not_from_backports(
             "firmware-linux-nonfree"
         )
-        if not update_firmware:
+        update_pipewire = check_package_not_from_backports("pipewire")
+        if not update_firmware and not update_pipewire:
             return
         add_backports_repo()
         # then pin firmware packages to backports repo
@@ -99,3 +109,6 @@ def bookworm_backports(os_data, log, **kwargs):
             with open(prefs_path, "w") as prefs:
                 if update_firmware:
                     prefs.write(prefs_firmware_data)
+
+                if update_pipewire:
+                    prefs.write(prefs_pipewire_data)
--- a/vmupdate/agent/source/plugins/bookworm_backports.py
+++ b/vmupdate/agent/source/plugins/bookworm_backports.py
@@ -1,0 +1,101 @@
+# coding=utf-8
+#
+# The Qubes OS Project, http://www.qubes-os.org
+#
+# Copyright (C) 2025  Marek Marczykowski-Górecki
+#                             <marmarek@invisiblethingslab.com>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+# USA.
+
+import os
+import subprocess
+
+sources_list = "/etc/apt/sources.list.d/backports.list"
+base_repo_url = "deb.debian.org/debian/dists/bookworm/InRelease"
+base_onion_repo_url = "2s4yqjx5ul6okpp3f2gaunr2syex5jgbfpfvhxxbbjwnrsvbk5v3qbid.onion/debian/dists/bookworm/InRelease"
+backports_line = (
+    "deb {baseurl} bookworm-backports main contrib non-free-firmware\n"
+)
+prefs_path = "/etc/apt/preferences.d/backports_pins"
+prefs_firmware_data = """\
+Package: src:firmware-nonfree
+Pin: release n=bookworm-backports
+Pin-Priority: 600
+
+"""
+
+
+def add_backports_repo():
+    # find URL flavor used for deb.debian.org
+    try:
+        output = subprocess.check_output(
+            ["apt-get", "--print-uris", "update"],
+            stderr=subprocess.DEVNULL,
+        )
+    except subprocess.CalledProcessError:
+        return
+    baseurl = None
+    backports_enabled = False
+    for url in output.decode().splitlines():
+        if not baseurl and (base_repo_url in url or base_onion_repo_url in url):
+            baseurl = (
+                url.split()[0]
+                .strip("'")
+                .replace("/dists/bookworm/InRelease", "")
+            )
+        if "/debian/dists/bookworm-backports/" in url:
+            # backports already enabled
+            backports_enabled = True
+    # add bookworm-backports if not already there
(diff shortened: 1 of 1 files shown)


mblanqui:
@marmarek, firmware-iwlwifi from testing is a far better fit, imo. firmware-iwlwifi from bookworm-backports is five months older and anecdotally I couldn't get it to work with my BE200 card (which NovaCustom also sells). Judging by the forum others have had problems with it too. https://forum.qubes-os.org/t/intel-wi-fi-7-be200-working/24017

marmarek:
AFAIK the "proper" Debian way of bringing packages from testing to stable is via backports, otherwise it isn't really safe to mix packages from testing and stable. While in this particular case it may work today, there isn't really guarantee it wouldn't break some other day (for example package in testing pulling some other dependencies from testing, that would break stable).
Anyway the last post in the thread you linked says the package from backports actually was new enough. If one day that wouldn't be the case anymore, the proper way is to request update via backports.

vmupdate/agent/source/plugins/firmware_backports.py
```diff
@@ -0,0 +1,64 @@
...
+            )
+            if b"bpo" in output:
+                # version from backports already installed
+                return
+        except subprocess.CalledProcessError:
+            return
+        # first, add bookworm-backports if not already there:
+        with open(sources_list) as sources:
+            current_sources = sources.read()
+        if "bookworm-backports" not in current_sources:
```
ben-grande, line 58: Whonix enables backports in `/etc/apt/sources.list.d/debian.list`.

It is best to check if it is enabled in any sources list, be it:

- sources.list
- sources.list.d/*.list
- sources.list.d/*.sources

Does not lead to an error, but a huge warning wall.

Quite difficult to validate this considering there are mirrors, protocols can vary (`https`, `http`, `tor+https`, `tor+http`, `http://HTTPS`) and URLs can also very, considering `debian.org`, `.onion` variant and mirrors.

A stricter validation would allow to not fail on false positives of external repositories having the suite `bookworm-backports`.

marmarek: Look for other deb.debian.org line and copy it? If not found, do nothing. The main target for this change is a template that may be used for sys-net. Without updating firmware from backports, https://github.com/QubesOS/qubes-issues/issues/9794 breaks wifi in most cases (for those using Debian for sys-net).

vmupdate/agent/source/plugins/firmware_backports.py
```diff
@@ -0,0 +1,64 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+# USA.
+
+import os
+import subprocess
+
+APT_CONF = "/etc/apt/apt.conf.d/01qubes-update"
+sources_list = "/etc/apt/sources.list"
+backports_line = "deb https://deb.debian.org/debian bookworm-backports main contrib non-free-firmware"
```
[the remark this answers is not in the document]
marmarek: Using new file in sources.list.d is a good idea. I can simply make it to not do anything if the file already exists (lets say, `backports.list`). apt-cacher-ng users can either adjust this file post-change, or proactively create it earlier with desired content.

vmupdate/agent/source/plugins/firmware_backports.py
```diff
@@ -0,0 +1,92 @@
...
+    if os_data.get("codename", "") == "bookworm":
+        # do anything only if firmware package is installed already
+        try:
+            output = subprocess.check_output(
+                ["dpkg", "-l", "firmware-linux-nonfree"],
+                stderr=subprocess.DEVNULL,
+            )
+            if b"bpo" in output:
+                # version from backports already installed
+                return
```
ben-grande, line 58: Now that pipewire was added, the check should not return as soon as firmware from backports is found.

vmupdate/agent/source/plugins/firmware_backports.py
ben-grande: I understand Whonix is not the focus, Kicksecure for example is becoming a template and there, as the base of Whonix, which could also become a sys-net template, has https://github.com/Kicksecure/anon-apt-sources-list/blob/master/etc/apt/sources.list.d/debian.list

Therefore I suggest testing against both of these URLs

- deb.debian.org/debian
- 2s4yqjx5ul6okpp3f2gaunr2syex5jgbfpfvhxxbbjwnrsvbk5v3qbid.onion/debian

vmupdate/agent/source/plugins/firmware_backports.py
ben-grande: Now I see why pipewire should be in a separate plugin, to avoid returning early, the plugin name is `firmware_backports` and the function has the same name.

Maybe switch the file name to `backports`, place backports related code in one function and separate the check for pipewire and firmware-linux-non-free?

ben-grande (commented):
Looks much more stricter now and `apt-get --print-uris` seems to be the best.

vmupdate/agent/source/plugins/firmware_backports.py
marmarek: I can also change this line to not return early.

vmupdate/agent/source/plugins/firmware_backports.py
marmarek: Is that an official onion repo? I don't see it listed at https://www.debian.org/mirror/list

ben-grande: Source: https://onion.debian.org/

Bottom of the page.

adrelanos:
I must caution against use of `/etc/apt/preferences.d/` / APT preferences / APT pinning. Please consider the following.

Quote https://www.kicksecure.com/wiki/Dev/APT_Pinning

> Debian developers would literally hate this.

Question: [Can a package pull a another package from another repository to be added?](https://lists.debian.org/debian-derivatives/2013/08/msg00004.html)

Answers:

* https://lists.debian.org/debian-derivatives/2013/08/msg00005.html
* https://lists.debian.org/debian-derivatives/2013/08/msg00006.html

> (And don't forget all the hate you will get from me and other deity@l.d.o inhabitants because your users end up reporting bugs about APT)

I rarely got such as direct request

marmarek:
So, back to the original question: how to mark a package to be installed from backports, similar to `apt-get -t bookworm-backports install firmware-linux-nonfree`  does, without installing the package yet? IIUC that apt-get call does mark the package somehow, so next updates continue to follow versions from backports, right? But I can't find how it's done, I tried diffing `/var/lib/{apt,dpkg}` before and after and don't see anything relevant...

ArrayBolt3:
The issues discussed in the threads @adrelanos linked to are mainly with *dynamically adding new sources.list and pin entries* during package installation or upgrade. Using pinning and the like is not inherently *bad*, it's just something that you should not be doing on already installed, running systems. People can and will customize their templates in all sorts of different ways, and adding a sources.list entry to their setup could cause major headaches if they have a non-standard setup.

The ideal way to do this is the way you're already doing - make sure the repos are right from the get-go on freshly built systems. But you need older systems to be updated as well so that users get fixes and don't suffer problems during a kernel upgrade.

I don't think there's any *good* solution here sadly - Debian dislikes the solution being proposed here, but I think it might be the best that can be done. Some solutions that I've thought of looking at this:

* Ideal but probably not practical solution: distribute the firmware-nonfree package from bookworm-backports in Qubes OS's repo. This means that no matter what setup a user has to use in order to get the packages from Qubes OS's repo, that setup will "just work" to pull the right package. Whether this is *legal* or not is another question, I am not a lawyer, nor do I know if Debian has special permissions here.
* Non-ideal solution requiring manual intervention: do something like what Kicksecure does and offer a script that, when run, adds the sources.list.d file. That way the user has to explicitly request its creation, and if things go south, they can fix whatever issues result.
* *Almost* practical solution: introduce the new sources.list file and pinning automatically, but use a debconf prompt to alert the user so they know what happened if something goes wrong. The problem with this is that most users are probably updating using Qubes Update, which means that they won't see this message.
* Overcomplicated and probably overly alarming solution: when the upgrade happens, have the upgraded Debian VMs send a qrexec request to dom0 that causes it to pop up a window with a message indicating that the backports repo is being enabled and this might cause apt breakage. dom0 would only ever show this prompt once to avoid flooding the user. This of course would be a huge pain to implement, and most users (all users who aren't messing with their apt config) aren't going to know what this means or what to do about it, thus support requests will come.
* Good enough solution: just carry on with this, if things break for some users, too bad for them. This is arguably a bad idea, but there are lots of packages out in the wild that do this sort of thing and... they work. Maybe for some fraction of users their systems break, but it's a rarity from what I know. This is even more true in Qubes OS - Debian has to support a myriad of different configurations and use cases, way more than Qubes OS has to support for its Debian-based VMs. What would be nice is if there was some way for users who do experience breakage to know that they should be reaching out to Qubes, not Debian, but thankfully for us Debian is hard enough to report a bug to and the Qubes OS forums are easily accessible enough that people will probably go to the forums first rather than fighting with `reportbug`.

PR #187, opened 2025-03-19
vmupdate: support dnf5 python API

piotrbartman:


vmupdate/agent/entrypoint.py
```diff
@@ -74,7 +74,14 @@ def get_package_manager(os_data, log, log_handler, log_level, no_progress):
...
+            version = int(os_data["release"].split(".")[0])
+        except ValueError:
+            version = 99  # fedora changed its version
+        try:
+            if version < 41:
+                from source.dnf.dnf_api import DNF as PackageManager
+            else:
+                from source.dnf.dnf5_api import DNF as PackageManager
         except ImportError:
             log.warning("Failed to load dnf with progress bar. Use dnf cli.")
```
marmarek, line 86: I hit this case initially. Several things:
- I didn't got this message in GUI updater, only in log file, why?
- The "use dnf cli" suggest an action from the user, -> "using"?
- Maybe try fallback to dnf4 before fallback to CLI? Or could that cause issues?
- Should core-agent-linux pull in python3-libdnf5? Or maybe qubes-vm-recommended (in which case, it won't be present in minimal template)? With DNF4 it wasn't an issue, as it was a dependency of DNF itself...

piotrbartman:
> But also, when keeping details open, qubes-update-gui gets quite busy during download phase - about 50% CPU all the time. It's about 20% when details are closed. And much lower regardless of viewing details when installing updates (after download completes). This may be an issue in qubes-update-gui, but probably related to the rate at which download progress is reported.

This is related to text output. I reduced the amount of  printed text, and it looks better. The GUI probably needs to be optimized in this regard.

vmupdate/agent/entrypoint.py
piotrbartman: - Should be better now
- fixed
- you're right
- I think it should

vmupdate/agent/source/dnf/dnf5_api.py
```diff
@@ -133,6 +133,49 @@ def __init__(self, weight: int, log):
...
+    ) -> int:
+        """
+        Download progress callback.
+
+        :param user_cb_data: Associated user data obtained from add_new_download.
+        :param total_to_download: Total number of bytes to download.
+        :param downloaded: Number of bytes downloaded.
+        """
+        if not self.fetching_notified:
+            print(f"Fetching {self.count} packages", flush=True)
```
marmarek, line 168: Maybe include total download size too here?

piotrbartman: Good idea, I did it for all package managers

marmarek:
Tests says repository refresh is reported as no updates instead of failure:
> Traceback (most recent call last):
  File "/usr/lib64/python3.13/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/vm_update.py", line 832, in test_130_no_network_qubes_vm_update
    self.update_via_proxy_qubes_vm_update_impl(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        method="qubes-vm-update",
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
        expect_updated=False,
        ^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/vm_update.py", line 706, in update_via_proxy_qubes_vm_update_impl
    self.run_qubes_vm_update_and_assert(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        expected_ret_codes=expected_ret_codes, options=options
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/vm_update.py", line 638, in run_qubes_vm_update_and_assert
    self.fail(
    ~~~~~~~~~^
        "qubes-vm-update return unexpected code: "
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        f"{proc.returncode} in {expected_ret_codes}\n"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        + f_log.read()
        ^^^^^^^^^^^^^^
    )
    ^
AssertionError: qubes-vm-update return unexpected code: 0 in (23,)

> test-inst-vm1 (pending):   0%|                                                                                                                | 0/100 [00:00<?, ?it/s]
test-inst-vm1 (updating):   0%|                                                                                                               | 0/100 [00:00<?, ?it/s]
test-inst-vm1 (no updates):   0%|                                                                                                             | 0/100 [00:02<?, ?it/s]
test-inst-vm1 (no updates):   0%|                                                                                                             | 0/100 [00:02<?, ?it/s]

vmupdate/agent/source/dnf/dnf5_api.py
```diff
@@ -0,0 +1,270 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+# USA.
+
+import subprocess
+import os
```
marmarek, line 24: Unused import?

marmarek:
Now I don't get what is currently being done at all... only "Fetching xxx packages" and then "Updating packages", and finally "Updated". My complain was only about "Fetching ..." message, as it looks like it was printed for all packages at the start, not when it actually started fetching that package - so not very useful. But "Fetched ..." message is meanigful as you see what was done already. And similar for installation - message which package is currently being installed (or just was installed) is an useful info.

As for CPU usage, even with this very minimal, it qubes-update-gui still used about 50% CPU during download, so it isn't (just) about messages. More likely frequency of updating the progress bar. Not sure what would be a good solution here, but it's a minor issue, can be solved later.

marmarek (approved):

PR #191, opened 2025-05-28
updater: dom0 non-iteractive updater

piotrbartman:
- `qubes-vm-update` uses `qubes-dom0-update --just-print-progress` for non-interactive update, only small subset of action are supported.
- introduce agent_type: for local (dom0), remote (template vms etc.) and proxy (update vm) actions. 
- progress reporting for dom0 update (dnf4, dnf5)

requires: QubesOS/qubes-core-agent-linux/pull/576


marmarek (changes requested):
This is based on initial reading the code, I haven't tested it yet.

marmarek:
I tried to play with it and I have some questions/remarks:

1. How can I make `qubes-dom0-update --targets=dom0` display the progress bar instead of full output?
2. When targeting dom0, it asks me to run qubes-vm-update as root. This usually leads to permission issues with log files, so we've been careful to always call it as normal user. Maybe qubes-vm-update should use sudo internally for updating dom0, instead of calling the whole qubes-vm-update as root from GUI?
3. Updating dom0 via GUI didn't worked for me (but via qubes-vm-update CLI worked). Fetching updates went fine, but then it took significantly too little time to apply updates (didn't actually install them) and the details section says:
```
...
Fetching lksctp-tools-0:1.0.20-1.fc41.x86_64 [99.44 KB]
Fetching lua-0:5.4.8-1.fc41.x86_64 [204.02 KB]
Fetching lua-libs-0:5.4.8-1.fc41.x86_64 [134.70 KB]
Fetching tzdata-java-0:2025b-1.fc41.noarch [47.40 KB]


No match for argument: *


                 1:24.8.7.2-2.fc41   qubes-dom0-cached
    1:21.0.7.0.6-2.fc41 qubes-dom0-cached
                      0.1.8-1.fc41        qubes-dom0-cached
                  0.1.12-1.fc41       qubes-dom0-cached
                 1:24.8.7.2-2.fc41   qubes-dom0-cached
            1:24.8.7.2-2.fc41   qubes-dom0-cached
            1:24.8.7.2-2.fc41   qubes-dom0-cached
            1:24.8.7.2-2.fc41   qubes-dom0-cached
            1:24.8.7.2-2.fc41   qubes-dom0-cached
            1:24.8.7.2-2.fc41   qubes-dom0-cached
      1:24.8.7.2-2.fc41   qubes-dom0-cached
   1:24.8.7.2-2.fc41   qubes-dom0-cached
            1:24.8.7.2-2.fc41   qubes-dom0-cached
         1:24.8.7.2-2.fc41   qubes-dom0-cached
         1:24.8.7.2-2.fc41   qubes-dom0-cached
     1:24.8.7.2-2.fc41   qubes-dom0-cached
            1:24.8.7.2-2.fc41   qubes-dom0-cached
        1:24.8.7.2-2.fc41   qubes-dom0-cached
1:24.8.7.2-2.fc41   qubes-dom0-cached
       1:24.8.7.2-2.fc41   qubes-dom0-cached
           1:24.8.7.2-2.fc41   qubes-dom0-cached
             1:24.8.7.2-2.fc41   qubes-dom0-cached
      1:24.8.7.2-2.fc41   qubes-dom0-cached
          1:24.8.7.2-2.fc41   qubes-dom0-cached
             1:24.8.7.2-2.fc41   qubes-dom0-cached
                    0.1.8-1.fc41        qubes-dom0-cached
                1.0.20-1.fc41       qubes-dom0-cached
                         5.4.8-1.fc41        qubes-dom0-cached
                    5.4.8-1.fc41        qubes-dom0-cached
                 2025b-1.fc41        qubes-dom0-cached
```

PS my approach to get some updates pending is this: `qubes-dom0-update --disablerepo=updates libreoffice`, and on latter attempts add `--action=downgrade`.

marmarek:
openQArun PR_LABEL=openqa-group-1 TEST=system_tests_network,system_tests_network_updates,system_tests_gui_tools MACHINE=64bit

marmarek:
openQArun PR_LABEL=openqa-group-1 TEST=system_tests_network,system_tests_network_updates,system_tests_gui_tools MACHINE=64bit

marmarek:
> * VmUpdates_fedora-42-xfce: [test_130_no_network_qubes_vm_update](https://openqa.qubes-os.org/tests/162439#step/VmUpdates_fedora-42-xfce/8) (failure)
>       `AssertionError: qubes-vm-update return unexpected code: 24 in (23,)`

Is it intentional change? It should fail (trying to update without network access), but it used to fail with different error code.

piotrbartman:
In dnf5, building the repo sack is more automated. When loading repositories, libdnf5 decides internally whether to refresh the data or if the cache is fresh enough (like dnf cli). This gives us less control over the metadata refresh step, even though it is required to build the repo sack object.

Currently, `refresh` in our wrapper of dnf5 simply invalidate cache ensures that the cache is updated with fresh data. However, skipping this step does not guarantee that a metadata refresh attempt won't happen anyway (since libdnf5 manages it).
We could force this behavior by modifying the config options to disable cache refreshing, but the question is whether we want or need to do that. 

RN the first try of network connection happens during the update step, which explains the different error code.

IMO we should update the test to accept both error codes for all package managers (as is already done for Arch), rather than modifying libdnf5 to mimic apt. Thoughts?

marmarek:
Fair enough.

marmarek:
openQArun PR_LABEL=openqa-group-1 TEST=system_tests_network,system_tests_network_updates,system_tests_gui_tools MACHINE=64bit

vmupdate/update_manager.py
```diff
@@ -121,21 +141,25 @@ def collect_result(self, result_tuple: Tuple[str, ProcessResult]):
...
+                self.print(FormatedLine(qube_name, "out", line))
+            for line in result.err.split("\n"):
+                self.print(FormatedLine(qube_name, "err", line))
         elif not self.quiet and self.no_progress:
             self.print(result.out)
 
     def print(self, *args):
         if self.buffered:
-            self.buffer += ' '.join(args) + '\n'
+            self.buffer += " ".join(str(args)) + "\n"
```
marmarek: Didn't you mean something like `map(str, args)`?

dom0-updates/qubes-dom0-update
```diff
@@ -310,29 +318,102 @@ qvm-run --nogui -q -- "$UPDATEVM" "rm -rf -- '$dom0_updates_dir/etc' '$dom0_upda
...
+        # will automatically sanitize them, but we explicitly tell it to anyway
+        # as a precaution.
+        #
+        # We MUST NOT use ‘exec script’ here.  That causes ‘script’ to
+        # inherit the child processes of the shell.  ‘script’ mishandles
+        # this and enters an infinite loop.
+        CMD="script --quiet --return --command '${CMD//\'/\'\\\'\'}' /dev/null"
+    fi
+    if [ "$PROGRESS_REPORTING" == "1" ]; then
+        qvm-run "${QVMRUN_OPTS[@]}" -- "$UPDATEVM" "$CMD" < /dev/null | sed "s/``^/$(hostname):out: /"
```
marmarek:
```suggestion
        qvm-run "${QVMRUN_OPTS[@]}" -- "$UPDATEVM" "$CMD" < /dev/null | sed "s/^/$(hostname):out: /"
```
?

marmarek (changes requested):
Besides two minor issues, error handling is broken:
```
dom0 (no updates):   0%|                                                                                                                                                                               | 0/100 [00:06<?, ?it/s]
( ' d o m 0 : e r r :   U s i n g   s y s - f i r e w a l l   a s   U p d a t e V M   f o r   D o m 0 ' , )
( ' d o m 0 : e r r :   D o w n l o a d i n g   u p d a t e s .   T h i s   m a y   t a k e   a   w h i l e . . . ' , )
( ' d o m 0 : e r r :   / b i n / b a s h :   l i n e   1 :   / u s r / l i b / q u b e s / q u b e s - d o w n l o a d - d o m 0 - u p d a t e s - i n i t . s h :   N o   s u c h   f i l e   o r   d i r e c t o r y ' , )
( ' d o m 0 : o u t :   [ 0 ; 3 1 m [ 0 m d o m 0 : o u t :   R e f r e s h i n g   p a c k a g e   i n f o ' , )
( ' d o m 0 : o u t :   s y s - f i r e w a l l : o u t :   ' , )
( ' d o m 0 : o u t :   s y s - f i r e w a l l : e r r :   ' , )
( ' d o m 0 : o u t :   d o m 0 : o u t :   N o t h i n g   t o   d o . ' , )
( ' d o m 0 : e r r :   d o m 0   d o n e   n o _ u p d a t e s ' , )
( < v m u p d a t e . a g e n t . s o u r c e . s t a t u s . F o r m a t e d L i n e   o b j e c t   a t   0 x 7 1 1 8 f 8 2 b 4 c 5 0 > , )
( < v m u p d a t e . a g e n t . s o u r c e . s t a t u s . F o r m a t e d L i n e   o b j e c t   a t   0 x 7 1 1 8 f 8 2 b 4 c 5 0 > , )

[user@dom0 ~]$ echo $?
0
```
failure of the script (here because it wasn't there, but applies to other cases too) should be reported as an error, not "no updates"

vmupdate/agent/source/pacman/pacman_cli.py
marmarek, line 1: pacman plugin needs an updated `__init__` arguments too

marmarek: oh, it needs more than that, `AttributeError: 'PACMANCLI' object has no attribute 'PROGRESS_REPORTING'`

dom0-updates/qubes-dom0-update
```diff
@@ -79,6 +81,12 @@ while [ $# -gt 0 ]; do
         --show-output)
             # ignore
             ;;
+        --quiet)
+            exec > /dev/null
+            ;;
```
marmarek: This breaks salt, see https://github.com/QubesOS/qubes-issues/issues/10543#issuecomment-3716492445

marmarek (changes requested):

dom0-updates/qubes-dom0-update
piotrbartman: I believe the easiest way would be to rename this flag, `--silent` ?

marmarek: Sounds good, `--silent` is not a valid dnf option, so should be safe.

marmarek (approved):
Approved as of 4824edd343b1bcbb4da1c0ecc421c886ae4d28fb

PR #193, opened 2025-06-05
add Zenity to dependencies

zpc0:
Some small programs uses Zenity to show dialogs.
Fixes: QubesOS/qubes-issues#7280

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -421,7 +421,11 @@ else
     fi
     echo "No updates available" >&2
     if [ "$GUI" == "1" ]; then
-        zenity --info --title='Dom0 updates' --text='No updates available'
+        if [ "$KDE_FULL_SESSION" ]; then
+            kdialog --msgbox 'No updates available'
+        else
+            zenity --info --title='Dom0 updates' --text='No updates available'
+        fi
     fi
 fi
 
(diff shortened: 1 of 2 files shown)


marmarek:
That may be a problem for KDE setups, which are supposed to use kdialog. I think it should be `(zenity or kdialog)`. And the single place that doesn't have fallback to kdialog should get one...

zpc0:
> I think it should be `(zenity or kdialog)`.

So adding `zenity/kdialog` to `qubes-linux-desktop-xfce/kde` instead of `core-admin-linux` maybe more clean approach?

zpc0:
I missed `Requires: (X if Y)` syntax...
Now this PR uses it.

marmarek:
> So adding `zenity/kdialog` to `qubes-linux-desktop-xfce/kde` instead of `core-admin-linux` maybe more clean approach?

Not really, it's this package that require them, and it need to work even if you don't have xfce or kde installed (for example some people use i3).

marmarek:
> I missed `Requires: (X if Y)` syntax...

It needs to work also without kde or xfce installed, that's why I asked for `or` operator, not `if`.

zpc0:
Thank you for pointing out.
Now `zenity` is pulled for non-XFCE and non-KDE systems.
On system without both `zenity` and `kdialog`, it is unclear whether `zenity` or `kdialog` is pulled if I use `or`.
But some script use `zenity` as default program, so I use `if` and `unless`.
Please tell me if I'm wrong.

marmarek (approved):

PR #194, opened 2025-06-16
vmupdate: answer "yes" to the import key question

marmarek:
When fetching repository metadata for the first time, dnf will ask for
confirmation on importing metadata signing key. Since the key is
imported from the local filesystem, it's safe to do, so do it
automatically.

QubesOS/qubes-issues#9807

### DIFF
--- a/vmupdate/agent/source/dnf/dnf_cli.py
+++ b/vmupdate/agent/source/dnf/dnf_cli.py
@@ -53,6 +53,7 @@ class DNFCLI(PackageManager):
         cmd = [self.package_manager,
                "-q",
                "check-update",
+               "--assumeyes",
                f"--setopt=skip_if_unavailable={int(not hard_fail)}"]
         result_check = self.run_cmd(cmd)
         # ret_code == 100 is not an error


marmarek:
All good questions.

> does this not have the potential to inadvertently accept remote keys at any upgrade?

Key location is specified in the repo file, all repos we include use file:// location. The same applies to all other popular repositories. If you modify the repo configuration to use remote location, then yes, it may import that key - you asked for it... But also, see the answer below.

> Also, `--assumeyes` has the potential to accept other questions with unwanted consequences?

Not really, there aren't any other prompts during refresh operation. In fact, this prompt is only about _metadata_ signatures (not package signatures), which most repositories don't use anyway...  

> Wouldn't it be more safe and robust to explicitly import the key from local filesystem when/if necessary?

If there would be a proper method for that, maybe (but also, it would require duplicating quite some part already done by dnf - for example to properly get repo config, including any drop-ins, check for enabled/disabled repos etc). But there isn't any API for importing metadata signing key - it's stored by dnf internally, keys imported to rpmdb via `rpm --import` are not the same thing. And the only "official" method is answering "yes" to that prompt...

marmarek:
> Well, I may be happy with the location but still want to manually monitor the effective fingerprint or other changes. We should also be prompted on any unexpected key changes in any case IMO.

Actually, I don't think so. The only way to get a different key in a local filesystem is that something (already authenticated) put it there. And making key rotation harder than it should be is a bad thing.

> Hmm. Would you like me to take a shot at opening an upstream issue?

It's an issue known for a long time: https://bugzilla.redhat.com/show_bug.cgi?id=1768206

PR #198, opened 2025-07-28
qubes-receive-updates: Remove arbitrary package name length requirement.

736-c41-2c1-e464fc974:
Removing arbitrary package length requirement (127) chars.
Rpm works just fine even with package names longer than 127 chars.

### DIFF
--- a/dom0-updates/qubes-receive-updates
+++ b/dom0-updates/qubes-receive-updates
@@ -38,7 +38,7 @@ comps_file = None
 if os.path.exists('/usr/share/qubes/Qubes-comps.xml'):
     comps_file = '/usr/share/qubes/Qubes-comps.xml'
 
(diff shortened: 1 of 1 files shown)


736-c41-2c1-e464fc974:
@marmarek thanks for your reply, in my cicd pipeline i do create dev-build artifacts which are suffixed by a rpm compatible name derived from the branch name.

Whilst `git` does not seem to have a max branch name length, github and gitlab both seem to enforce 255 bytes.
So to answer your question: I'd need at least 255 ascii chars.

736-c41-2c1-e464fc974:
But, if we're worried about static buffer corner cases, we should maybe set this to the old documented rpm limit (65).
Not `127` or higher. So struggle a bit to understand that argument.

736-c41-2c1-e464fc974:
That's good to know, thanks @marmarek !
Updating the pr now.

marmarek (approved):

PR #200, opened 2025-09-05
Add flatpak update plugin for system-wide flatpaks

alimirjamali:
improves: https://github.com/QubesOS/qubes-issues/issues/2766
requires: https://github.com/QubesOS/qubes-core-agent-linux/pull/604

### DIFF
--- a/vmupdate/agent/source/plugins/flatpak_updater.py
+++ b/vmupdate/agent/source/plugins/flatpak_updater.py
@@ -1,0 +1,43 @@
+# coding=utf-8
+#
+# The Qubes OS Project, http://www.qubes-os.org
+#
+# Copyright (C) 2025  Ali Mirjamali <ali@mirjamali.com>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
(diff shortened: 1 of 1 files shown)


vmupdate/agent/source/plugins/flatpak_updater.py
```diff
@@ -0,0 +1,43 @@
...
+import subprocess
+
+from pathlib import Path
+
+def flatpak_updater(os_data, log, **kwargs):
+    """
+    Distro agnostic plugin to update system-wide flatpaks
+    """
+
+    if not Path("/usr/bin/flatpak"):
```
marmarek, line 31: Did you mean `.exists()`? but then, I don't think you can assume it's always `/usr/bin`, especially with NixOS existing out there...
We use `distutils.spawn.find_executable()` in various places

vmupdate/agent/source/plugins/flatpak_updater.py
```diff
@@ -0,0 +1,43 @@
...
+        return
+    log.info("Flatpak is installed. Checking for flatpak updates.")
+    if not subprocess.Popen(
+        ["flatpak", "remote-ls", "--system", "--updates"],
+        stdout=subprocess.PIPE
+    ).stdout.read().decode().strip():
```
marmarek, line 37: This should consider exit code (and report as update failure, not "no updates". `subprocess.check_output` may be helpful.

marmarek (changes requested):
Besides the two specific comments below, this should be a bit better integrated. At the very least regarding outcome reporting - it does make a difference if there were some updates or not (for example, if there were no updates, updater will not propose restarting relevant app qubes). Right now update plugins don't return anything (implicit `None`). Maybe consider return value similar to normal updater if a plugin returns not `None`?

vmupdate/agent/source/plugins/flatpak_updater.py
alimirjamali: > We use `distutils.spawn.find_executable()` in various places

Where? `distutils` is [deprecated](https://docs.python.org/3/library/distutils.html) since version 3.10, removed in version 3.12.

Anyways, I will replace it with appropriate supported lib (i.e. `shutil.which`).

PR #201, opened 2025-09-07
Pass `releasever` to DNF in UpdateVM

### LINKED ISSUE #7229: Debian-based updatevm show "unable to detect release version" warning
### Qubes OS release

4.1-rc3

### Brief summary

This warning is a bit scary. Not sure what it implies, but if it's harmless it would surely be better to get it silenced.

alimirjamali:
Avoid "unable to detect release version" messages

resolves: https://github.com/QubesOS/qubes-issues/issues/7229

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -312,6 +312,9 @@ qvm-run --nogui -q -- "$UPDATEVM" "rm -rf -- '$dom0_updates_dir/etc' '$dom0_upda
 
 CMD="/usr/lib/qubes/qubes-download-dom0-updates.sh --doit --nogui"
 
+# Properly pass dom0 qubes version to DNF in UpdateVM
+CMD="$CMD --releasever=$(awk -F'=' '/VERSION_ID/ {print $2}' /etc/os-release)"
+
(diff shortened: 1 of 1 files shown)


marmarek:
I don't think this is a good idea. The "unable to detect release version" message means DNF didn't found `qubes-release` package version, which likely means it didn't found the whole rpmdb. Passing this option will only hide the symptom, but not solve the actual issue.
The actual issue is likely related to Debian expecting different rpmdb path (in user home...). There is a bit of code handling that in the core-agent-linux part, but maybe it doesn't cover all the dnf calls? Or maybe it's broken in some cases?

alimirjamali:
> I don't think this is a good idea. The "unable to detect release version" message means DNF didn't found `qubes-release` package version, which likely means it didn't found the whole rpmdb. Passing this option will only hide the symptom, but not solve the actual issue. The actual issue is likely related to Debian expecting different rpmdb path (in user home...). There is a bit of code handling that in the core-agent-linux part, but maybe it doesn't cover all the dnf calls? Or maybe it's broken in some cases?

Ok. I will look at `qubes-core-agent-linux` and see what I can do there

alimirjamali:
Closing if favour of https://github.com/QubesOS/qubes-core-agent-linux/pull/605

PR #203, opened 2025-10-26
Enable only one display manager

marmarek:
They all have `Alias=display-manager.service` in the unit file, and
having more of them enabled at the same time causes conflicts (for
example `systemctl preset` call during installation fails, same as
`systemctl preset-all` on release upgrade.

### DIFF
--- a/system-config/75-qubes-dom0.preset
+++ b/system-config/75-qubes-dom0.preset
@@ -1,10 +1,9 @@
-enable gdm.service
 enable lightdm.service
-enable slim.service
-enable lxdm.service
-enable sddm.service
-enable kdm.service
-enable xdm.service
+#enable slim.service
+#enable lxdm.service
+#enable sddm.service
+#enable kdm.service
+#enable xdm.service
 
 
 disable systemd-timesyncd.service


marmarek:
@fepitre @HW42 @ben-grande is this change a good idea? I made it just to avoid `systemctl preset-all` failing, but there is also an alternative solution: https://github.com/QubesOS/qubes-dist-upgrade/pull/22
On the other hand, this might break cases when the user uninstalls lightdm to use something else. What do you think?

ben-grande:
> @fepitre @HW42 @ben-grande is this change a good idea? I made it just to avoid `systemctl preset-all` failing, but there is also an alternative solution: [QubesOS/qubes-dist-upgrade#22](https://github.com/QubesOS/qubes-dist-upgrade/pull/22) On the other hand, this might break cases when the user uninstalls lightdm to use something else. What do you think?

I think that not enabling via preset the login managers by default is better than hiding errors of `preset-all`.

If they uninstall `lightdm` to use something else, they need to enable something else as per Fedora's policy of services being disabled by default. It is true that it might fail in cases the user relies on the other display/login manager being enabled by default, but [Qubes' KDE documentation](https://doc.qubes-os.org/en/latest/user/advanced-topics/kde.html#login-manager) already documents disabling `lightdm` and enabling `sddm`.

If we don't want to break user upgrades, makes sense to handle the `preset-all` failure, but at least OpenQA should test without the handling, which would create a non-user tested case, which is not nice for finding bugs...

marmarek:
The problem is that release upgrade calls `preset-all`, which is important to enable new services (we can't really enforce new preset file to be installed before all the other packages, too many dependency cycles...). This applies both to qubes services (for example change to libvirt services structure - now they are modular), and to fedora services (not sure if there are significant changes fc37 -> fc41, but for example before it was needed for dbus-daemon -> dbus-broker migration, otherwise you end up without dbus daemon at all and everything breaks...). 

So, if calling `preset-all` on release upgrade, with this PR it would disable other display managers, that user might have enabled. Maybe upgrade should record which one was enabled before and re-enable it later?
Alternatively, instead of calling `preset-all`, the upgrade script could preset just selected units we know about (based on diff in preset files between R4.2 and R4.3?). But that feels fragile...

ben-grande:
> So, if calling `preset-all` on release upgrade, with this PR it would disable other display managers, that user might have enabled. Maybe upgrade should record which one was enabled before and re-enable it later?

I see now... yes, it should record, but doing that for every release upgrade for every relevant package? What about sshd not being enabled via preset.... It should be documented that user must add preset policies in dom0.

For the units that we do know, use [ignore directive](https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html)?

ben-grande:
> For the units that we do know, use [ignore directive](https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html)?

Mostly ok except it will enable `lightdm` and keep enabled `sddm`? In that case, something else need to detect that there is a display manager enabled to not enable lightdm....

ben-grande:
> For the units that we do know, use [ignore directive](https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html)?

Yeah, this still requires handling errors on `preset-all` because it will fail to enabled lightdm if another display manager is enabled. Maybe also `ignore lightdm`?

> > For the units that we do know, use [ignore directive](https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html)?
> 
> Mostly ok except it will enable `lightdm` and keep enabled `sddm`? In that case, something else need to detect that there is a display manager enabled to not enable lightdm....

```
systemctl disable lightdm; systemctl enable sddm
systemctl is-enabled lightdm sddm
> disabled
> enabled
```
```
# preset
ignore lightdm.service
ignore sddm.service
```
No errors on `systemctl preset-all`.

```sh
systemctl preset-all
systemctl is-enabled lightdm sddm
> disabled
> enabled
```

marmarek:
Ok, then `ignore` looks like a good option - including for `lightdm`. And then have something else enable it on the initial installation.

PR #207, opened 2026-02-04
Replace `fgrep` with `grep -F` to avoid a warning

mattmccutchen:
Today, I updated the dom0 kernel of my Qubes 4.3 system and saw a bunch of "warning: fgrep is obsolescent; using grep -F" messages.  This change (if backported) should stop those messages, FWIW.  Untested but hopefully obviously correct.

### DIFF
--- a/system-config/kernel-xen-efi.install
+++ b/system-config/kernel-xen-efi.install
@@ -38,7 +38,7 @@ case "$COMMAND" in
             dracut -f "$EFI_DIR/initramfs-${KVER}.img" "$KVER"
         fi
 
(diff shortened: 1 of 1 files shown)


marmarek (approved):
The change itself is okay, but it seems there is still `fgrep` call somewhere else - the warning is shown during `grub2-mkconfig` call.

mattmccutchen:
I think you are right and the `fgrep` call I removed is not the one that was causing the warnings on my system.  That's what I get for not testing the change.  Oh well; thanks for your patience.  I think QubesOS/qubes-issues#10536 has some information about the real cause of the warnings.

PR #208, opened 2026-03-02
vmupdate: wait for full VM shutdown before next queued update

### LINKED ISSUE #10617: Qubes update cannot handle hvms with attached pci devices

### Qubes OS release

4.2.4

### Brief summary

Qubes update GUI (with concurrency set to 1) and `qubes-vm-update --max-concurrency=1` are unable to handle multiple hvms that operate at least one same pcie device. Next qube in the queue is started without waiting for previous one to shut down completely, which causes pcie device conflicts.

### Steps to reproduce

1. Create at least two hvms that have at least one same pcie device attached to all of them
2. Attempt to update all of them
    - Set concurrency to 1 for this step

### Expected behavior

If currently updated vm has at least one same pcie device attached to it as the next vm in queue, wait for complete shutdown before starting the next vm.

### Actual behavior

Updating mechanism attempts to boot the second hvm immediately after initiating shutdown of the first hvm, which results in a pcie device conflict.

### Additional information

This might affect 4.3 as well, I haven't tested it there.

It would be nice to see this fixed for higher concurrency settings as well :)

Jayant-kernel:
### Summary
Fix update sequencing for VMs started by `qubes-vm-update`: wait for full shutdown in `QubeConnection.__exit__()` before returning.

This prevents the next queued VM from being started while the previous one is still in the shutdown phase and still holding shared PCI devices.

Closes QubesOS/qubes-issues#10617.

### What changed
- `vmupdate/qube_connection.py`
- After calling `self.qube.shutdown()` for VMs started by updater, call `wait_for_domain_shutdown([self.qube])`.

### Why
With `--max-concurrency=1`, the updater previously issued shutdown and immediately returned, allowing the next VM to start too early. For HVMs sharing PCI devices, this causes startup conflicts.

### Tests
- Added: `vmupdate/tests/test_qube_connection.py`
- `test_wait_for_shutdown_when_vm_started_by_update`
- `test_do_not_shutdown_if_vm_was_already_running`


Jayant-kernel:
Hi @piotrbartman @marmarek , quick update from my side: I pushed the fix for #10617 and updated the PR.
When you get a moment, could you please take a look?
Happy to tweak anything needed.

ben-grande:
It only needs to wait shutdown if it is a qube with PCI devices assigned to it. Making it wait for every qube to shutdown, slows the process. Also, no need to tag maintainers, they get notified if they are tracking the repository.

Jayant-kernel:
Yes, I updated the PR to wait only for VMs with assigned PCI devices, not for every VM shutdown. Please take another look when you can.

Jayant-kernel:
I uploaded my signing key to the keyserver and force-pushed the signed commit. 
 Could someone please re-run the signature check on this PR when possible? Thanks!

ben-grande:
Please remove this commit: https://github.com/QubesOS/qubes-core-admin-linux/pull/208/commits/c7eb4b2c3f07670aa45719a31ff8b80854476c31

vmupdate/tests/test_qube_connection.py
```diff
@@ -0,0 +1,77 @@
+# coding=utf-8
+#
+# The Qubes OS Project, https://www.qubes-os.org
+#
+# Copyright (C) 2026
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
```
ben-grande, line 10: This verbatim license seems to be too short compared to other files. I have not seen such such option being used.

https://github.com/QubesOS/qubes-core-admin-linux/blob/main/vmupdate/tests/test_vmupdate.py

vmupdate/tests/test_qube_connection.py
```diff
@@ -0,0 +1,77 @@
...
+#
+from unittest.mock import Mock, patch
+
+from vmupdate.qube_connection import QubeConnection
+
+
+@patch("vmupdate.qube_connection.wait_for_domain_shutdown")
+@patch("vmupdate.qube_connection.asyncio.run")
+def test_wait_for_shutdown_when_vm_started_by_update(_arun, wait_shutdown):
```
ben-grande: I don't think the `_` should be used, it normally is for options that you won't use it later, but would like to know on definition what they are. You use this option later:

```
 _arun.assert_called_once()
```

vmupdate/tests/test_qube_connection.py
```diff
@@ -0,0 +1,77 @@
...
+        pass
+
+    vm.shutdown.assert_called_once_with()
+    wait_shutdown.assert_called_once_with([vm])
+    _arun.assert_called_once()
+
+
+@patch("vmupdate.qube_connection.wait_for_domain_shutdown")
+@patch("vmupdate.qube_connection.asyncio.run")
+def test_do_not_wait_for_shutdown_without_assigned_pci(_arun, wait_shutdown):
```
ben-grande: https://github.com/QubesOS/qubes-core-admin-linux/pull/208/changes#r2882821840

vmupdate/tests/test_qube_connection.py
```diff
@@ -0,0 +1,77 @@
...
+        pass
+
+    vm.shutdown.assert_called_once_with()
+    wait_shutdown.assert_not_called()
+    _arun.assert_not_called()
+
+
+@patch("vmupdate.qube_connection.wait_for_domain_shutdown")
+@patch("vmupdate.qube_connection.asyncio.run")
+def test_do_not_shutdown_if_vm_was_already_running(_arun, wait_shutdown):
```
ben-grande: https://github.com/QubesOS/qubes-core-admin-linux/pull/208/changes#r2882821840

vmupdate/qube_connection.py
```diff
@@ -94,9 +96,19 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         if self.qube.is_running() and not self._initially_running:
             self.logger.info('Shutdown %s', self.qube.name)
```
ben-grande: It would be nice if the log message could be improved, add some text related to waiting shutdown, to differentiate from the default case that doesn't wait.

vmupdate/qube_connection.py
```diff
@@ -94,9 +96,19 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
             self.qube.shutdown()
+            if self._has_assigned_pci_devices(self.qube):
+                asyncio.run(wait_for_domain_shutdown([self.qube]))
```
ben-grande: Use `vmupdate.shutdown_domains()` instead.

vmupdate/qube_connection.py
```diff
@@ -94,9 +96,19 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
         self.__connected = False
 
+    @staticmethod
+    def _has_assigned_pci_devices(vm) -> bool:
+        """Return True when VM has assigned PCI devices."""
+        try:
+            return any(vm.devices['pci'].get_assigned_devices())
+        except Exception:  # best effort check only
```
ben-grande: This exception is too broad. Use `QubesDaemonAccessErrror`. Other exception should be raised.

Jayant-kernel:
Hey @ben-grande 
I have

Fixed the license header to match other test files
Renamed _arun to arun in the tests
Switched to shutdown_domains() instead of calling asyncio.run directly
Narrowed the exception to QubesDaemonAccessError
Improved the log message to mention PCI devices

Is this okay?

vmupdate/qube_connection.py
```diff
@@ -92,11 +93,26 @@ def __exit__(self, exc_type, exc_val, exc_tb):
                                   self.dest_dir, str(err))
 
         if self.qube.is_running() and not self._initially_running:
-            self.logger.info('Shutdown %s', self.qube.name)
-            self.qube.shutdown()
+            if self._has_assigned_pci_devices(self.qube):
+                self.logger.info(
+                    'Shutdown %s and wait for full shutdown '
```
ben-grande: Instead of saying "shutdown" twice, write: `Waiting for full shutdown %s (PCI devices assigned)`

vmupdate/qube_connection.py
```diff
@@ -92,11 +93,26 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                    '(PCI devices assigned)', self.qube.name)
+                from vmupdate.vmupdate import shutdown_domains
```
ben-grande: Please don't import it here, import at the top of the module.

vmupdate/qube_connection.py
Jayant-kernel: changed to 'Waiting for full shutdown %s (PCI devices assigned)

vmupdate/qube_connection.py
Jayant-kernel: moved shutdown_domains import to the top of the module.

Jayant-kernel:
@ben-grande 
rereview it

ben-grande:
CI is failing:

- https://gitlab.com/QubesOS/qubes-core-admin-linux/-/jobs/13389347545

Please wait for CI to complete before asking for review.

vmupdate/qube_connection.py
```diff
@@ -92,11 +93,28 @@ def __exit__(self, exc_type, exc_val, exc_tb):
                                   self.dest_dir, str(err))
 
         if self.qube.is_running() and not self._initially_running:
-            self.logger.info('Shutdown %s', self.qube.name)
-            self.qube.shutdown()
+            if self._has_assigned_pci_devices(self.qube):
+                # Late import to avoid circular dependency:
+                # qube_connection <- vmupdate <- update_manager <- qube_connection
+                from vmupdate.vmupdate import shutdown_domains  # pylint: disable=import-outside-toplevel
```
ben-grande: To avoid circular import, use a third file that holds the useful functions from vmupdate.py.

Jayant-kernel: I have  refactor it by moving the shared functions to a separate file to avoid the circular import.

ben-grande: Not that single function, all useful functions. It appears weird to have a module for holding a single function when vmupdate has multiple reusable functions. Although it fixes the current need, it is best to be future-proof in this case. Read the `vmupdate/vmupdate.py` and check which functions you can extract from there.

Order the commits like this:

1. Move useful (reusable) functions to utils module
2. Make the vm shutdown handle pci devices

ben-grande: Please read the above again. If you don't understand something, please ask.

Jayant-kernel:
@ben-grande @marmarek
PLease review it again

ben-grande:
Did not find errors on OpenQA related to this PR.

marmarek:
I wanted to say you could use `vm.shutdown(wait=True)` to wait for a single qube (the `shutdown_domains` helper function is useful to shutdown many in parallel). But, it seems the shutdown function currently doesn't expose the wait parameter... So, it's okay to leave as is now.

marmarek (approved):

ben-grande:
> But, it seems the shutdown function currently doesn't expose the wait parameter

https://github.com/QubesOS/qubes-core-admin-client/pull/469

PR #209, opened 2026-04-26
validate env limits in qfile-dom0-unpacker

### LINKED ISSUE #8882: Make qfile-dom0-unpacker validate values set via environment variables

I want to be able to use qfile-agent in dom0 and when dealing with large files, I need to modify `UPDATES_MAX_BYTES` and `UPDATES_MAX_FILES`. Unfortunately, it [does not validate](https://github.com/QubesOS/qubes-core-admin-linux/blob/main/dom0-updates/qfile-dom0-unpacker.c#L91-L94) the variables. If the variable is invalid, `atoll()` will convert it to `0`, thus no limit is set instead of failing.

### The solution you'd like

Fail when the user set a value different than zero and `bytes_limit` or `files_limit` is converted to `0`.

### The value to a user, and who that user might be

In order to complete
- https://github.com/QubesOS/qubes-issues/issues/3633#issuecomment-1900624266

It must fail on invalid user input instead of allowing the user mistake to unlimit the values.

rishi-jat:
### Summary 

This change fixes unsafe parsing of UPDATES_MAX_BYTES and UPDATES_MAX_FILES in qfile-dom0-unpacker.

The current implementation uses atoll(), which silently returns 0 on invalid input. Since 0 is interpreted as “no limit”, malformed or unintended values (e.g., non-numeric strings) can effectively disable limits without any error. In the context of copying data into dom0, this results in a fail-open behavior and weakens expected safeguards.

This patch replaces atoll() with strtoll() and adds strict validation. The parsing now:
- rejects non-numeric and malformed inputs
- rejects partial parses (e.g., trailing characters)
- rejects overflow conditions (ERANGE)
- rejects negative values
- accepts 0 only when explicitly provided as "0"

Any invalid input results in an error message and immediate exit, enforcing fail-closed semantics.

This ensures that user misconfiguration cannot silently remove limits and aligns the behavior with the expectation that invalid input must not degrade safety guarantees.

Fixes QubesOS/qubes-issues#8882

### DIFF
--- a/dom0-updates/Makefile
+++ b/dom0-updates/Makefile
@@ -1,4 +1,4 @@
 CC=gcc
 CFLAGS=-g -I. -Wall -Wextra -Werror -fPIC -pie
 qfile-dom0-unpacker: qfile-dom0-unpacker.o
-	$(CC) -pie -g -o $@ $^ -lqubes-rpc-filecopy
+	$(CC) -pie -g -o $@ $^ -lqubes-rpc-filecopy -lqubes-pure
--- a/dom0-updates/qfile-dom0-unpacker.c
+++ b/dom0-updates/qfile-dom0-unpacker.c
@@ -19,6 +19,32 @@
 #define min(a,b) ((a) < (b) ? (a) : (b))
 #define max(a,b) ((a) > (b) ? (a) : (b))
 
+static long long parse_limit_env(const char *name, long long fallback)
+{
+    const char *value = getenv(name);
+    if (!value)
+        return fallback;
+
+    if (*value < '0' || *value > '9') {
+        fprintf(stderr, "Invalid value for %s: %s\n", name, value);
+        exit(1);
+    }
+
+    errno = 0;
+    char *end = NULL;
+    long long limit = strtoll(value, &end, 10);
+    if (errno == ERANGE || *end != '\0' || limit < 0) {
+        fprintf(stderr, "Invalid value for %s: %s\n", name, value);
+        exit(1);
+    }
+    if (limit == 0 && strcmp(value, "0") != 0) {
+        fprintf(stderr, "Invalid value for %s: %s\n", name, value);
+        exit(1);
+    }
+
+    return limit;
+}
+
 int prepare_creds_return_uid(const char *username)
 {
     struct passwd *pwd;
@@ -88,10 +114,8 @@ int main(int argc, char ** argv)
         perror("Failed to check free space");
     }
 
-    if ((var=getenv("UPDATES_MAX_BYTES")))
-        bytes_limit = atoll(var);
-    if ((var=getenv("UPDATES_MAX_FILES")))
-        files_limit = atoll(var);
+    bytes_limit = parse_limit_env("UPDATES_MAX_BYTES", bytes_limit);
+    files_limit = parse_limit_env("UPDATES_MAX_FILES", files_limit);
 
     set_size_limit(bytes_limit, files_limit);
 
--- a/dom0-updates/qfile-dom0-unpacker.c
+++ b/dom0-updates/qfile-dom0-unpacker.c
@@ -26,19 +26,23 @@ static long long parse_limit_env(const char *name, long long fallback)
         return fallback;
 
     if (*value < '0' || *value > '9') {
-        fprintf(stderr, "Invalid value for %s: %s\n", name, value);
+        fprintf(stderr, "Invalid value for %s: %s: not a non-negative integer\n", name, value);
         exit(1);
     }
 
     errno = 0;
     char *end = NULL;
     long long limit = strtoll(value, &end, 10);
-    if (errno == ERANGE || *end != '\0' || limit < 0) {
-        fprintf(stderr, "Invalid value for %s: %s\n", name, value);
+    if (errno == ERANGE) {
+        fprintf(stderr, "Invalid value for %s: %s: out of range\n", name, value);
+        exit(1);
+    }
+    if (*end != '\0') {
+        fprintf(stderr, "Invalid value for %s: %s: trailing non-numeric characters\n", name, value);
         exit(1);
     }
     if (limit == 0 && strcmp(value, "0") != 0) {
-        fprintf(stderr, "Invalid value for %s: %s\n", name, value);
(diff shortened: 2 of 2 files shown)


DemiMarie:
@rishi-jat If this was assisted by AI you need to mention that it was.  See the Qubes OS documentation.

(no longer a team member, but familiar with this rule)

ben-grande:
I am no C expert, I just tested the code. Error messages should contain the reason of the failure, to be able to differentiate amongst them. The last if statement seems to be never reached as errors fall under the previous conditions, but it might be ok to keep it.

ben-grande:
Didn't find failures related to this PR. Because this doesn't feel urgent, we can wait for the next openqarun that will have less failures to better see whats important, as https://github.com/QubesOS/qubes-core-admin/pull/804 has been fixed.

ben-grande:
This seems similar to https://github.com/QubesOS/qubes-core-qrexec/blob/f6e44eaad81e1eb2465e017b875ae36a03e42277/daemon/qrexec-client.c#L118. Maybe it should be a shared library.

Maybe here: https://github.com/QubesOS/qubes-linux-utils/tree/main/qrexec-lib.

Wait for Marek to confirm becore making any changes.

marmarek:
> This seems similar to https://github.com/QubesOS/qubes-core-qrexec/blob/f6e44eaad81e1eb2465e017b875ae36a03e42277/daemon/qrexec-client.c#L118. Maybe it should be a shared library.
> 
> Maybe here: https://github.com/QubesOS/qubes-linux-utils/tree/main/qrexec-lib.
> 
> Wait for Marek to confirm becore making any changes.

Yes, good idea. There are a bunch of helper functions in pure.c/pure.h already, so that would be a good place for this one too.

marmarek:
@rishi-jat are you still working on this?

rishi-jat:
> @rishi-jat are you still working on this?

Sorry for the delay sir , forgot about this issue/PR -- the PR is my queue will update asap. Thanks!

rishi-jat:
@marmarek PTAL. Thanks!

ben-grande:
> > This seems similar to https://github.com/QubesOS/qubes-core-qrexec/blob/f6e44eaad81e1eb2465e017b875ae36a03e42277/daemon/qrexec-client.c#L118. Maybe it should be a shared library.
> > Maybe here: https://github.com/QubesOS/qubes-linux-utils/tree/main/qrexec-lib.
> > Wait for Marek to confirm becore making any changes.
> 
> Yes, good idea. There are a bunch of helper functions in pure.c/pure.h already, so that would be a good place for this one too.

@rishi-jat please see this comment above.

marmarek:
On Tue, Jun 16, 2026 at 04:36:16PM -0700, bhavishyaa-sahuu wrote:
> bhavishyaa-sahuu left a comment (QubesOS/qubes-core-admin-linux#209)

For some reason I don't see this comment on the github UI...

> Applied maintainer feedback:
> 
> - Improved error messages with specific reasons (already in previous commit).
> - Removed the unreachable `if (limit < 0)` branch in `parse_limit_env` (it could never be hit due to the leading-digit check before `strtoll`; this addresses the note that some conditions seemed never reached).
> - Cleaned up branch history: rebased the two feature commits cleanly on top of current `main` (the branch previously carried unrelated merge history from other PRs). PR diff is now minimal and only touches the intended file.
> 
> Regarding the shared library suggestion (https://github.com/QubesOS/qubes-core-admin-linux/pull/209#issuecomment-4374958656 and follow-ups): a general helper (similar to `parse_int` in qrexec or the pure validation funcs) belongs in `qubes-linux-utils/qrexec-lib/pure.h` + impl (as @marmarek suggested). This PR keeps a small static helper local to the unpacker because:
> - the function does I/O + exit on error (getenv, fprintf+exit), while pure/ is for side-effect free validators
> - it can be extracted/moved later without API break once a shared `qubes_pure_strtonum` / `qubes_parse_limit` etc. is introduced and the `qubes-utils-devel` BuildRequires is bumped.
> 
> This should be ready to merge (the functional change is isolated, tested for error paths, and history clean). OpenQA etc. can be re-run as needed.

Currently this fails to build, looks like the PR to linux-utils package
is missing (with the qubes_pure_parse_nonneg_ll function).

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

rishi-jat:
> On Tue, Jun 16, 2026 at 04:36:16PM -0700, bhavishyaa-sahuu wrote:
> > bhavishyaa-sahuu left a comment (QubesOS/qubes-core-admin-linux#209)
> 
> For some reason I don't see this comment on the github UI...
> 
> > Applied maintainer feedback:
> > 
> > - Improved error messages with specific reasons (already in previous commit).
> > - Removed the unreachable `if (limit < 0)` branch in `parse_limit_env` (it could never be hit due to the leading-digit check before `strtoll`; this addresses the note that some conditions seemed never reached).
> > - Cleaned up branch history: rebased the two feature commits cleanly on top of current `main` (the branch previously carried unrelated merge history from other PRs). PR diff is now minimal and only touches the intended file.
> > 
> > Regarding the shared library suggestion (https://github.com/QubesOS/qubes-core-admin-linux/pull/209#issuecomment-4374958656 and follow-ups): a general helper (similar to `parse_int` in qrexec or the pure validation funcs) belongs in `qubes-linux-utils/qrexec-lib/pure.h` + impl (as @marmarek suggested). This PR keeps a small static helper local to the unpacker because:
> > - the function does I/O + exit on error (getenv, fprintf+exit), while pure/ is for side-effect free validators
> > - it can be extracted/moved later without API break once a shared `qubes_pure_strtonum` / `qubes_parse_limit` etc. is introduced and the `qubes-utils-devel` BuildRequires is bumped.
> > 
> > This should be ready to merge (the functional change is isolated, tested for error paths, and history clean). OpenQA etc. can be re-run as needed.
> 
> Currently this fails to build, looks like the PR to linux-utils package
> is missing (with the qubes_pure_parse_nonneg_ll function).
> 
> -- 
> Best Regards,
> Marek Marczykowski-Górecki
> Invisible Things Lab
> 

Yep i also seen that message on mail but not able to see on GitHub Ui -- I haven't raised PR to Linux-utils package yet will open asap. Thanks!

PR #212, opened 2026-05-22
Use admin utils wrappers for vm actions

ben-grande:
Please see: https://github.com/QubesOS/qubes-core-admin-client/pull/469#issuecomment-4518848847

---

Didn't test locally, only know that unit tests pass.

### DIFF
--- a/vmupdate/qube_connection.py
+++ b/vmupdate/qube_connection.py
@@ -20,6 +20,7 @@
 # USA.
 
 import argparse
+import asyncio
 import os
 import shutil
 import signal
@@ -34,11 +35,11 @@ from typing import List, Self, Any, Type
 import qubesadmin
 import qubesadmin.exc
 from qubesadmin.vm import QubesVM
+from qubesadmin.utils import shutdown
 from vmupdate.agent.source.args import AgentArgs
 from vmupdate.agent.source.log_config import LOGPATH, LOG_FILE
 from vmupdate.agent.source.status import StatusInfo, FinalStatus, FormatedLine
 from vmupdate.agent.source.common.process_result import ProcessResult
-from vmupdate.utils import shutdown_domains
 
 
 class QubeConnection:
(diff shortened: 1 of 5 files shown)


ben-grande:
Still didn't test locally, but the main PR to core-admin-client and another very similar PR https://github.com/QubesOS/qubes-desktop-linux-manager/pull/312/changes/47565a510ad13061c0ca80981273eb10ea11af3d, were tested by openqa on https://openqa.qubes-os.org/tests/180820, and they did pass, so marking this PR as ready for review.

vmupdate/qube_connection.py
```diff
@@ -93,14 +95,13 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
             if self._has_assigned_pci_devices(self.qube):
                 self.logger.info(
                     'Waiting for full shutdown %s (PCI devices assigned)',
                     self.qube.name)
-                shutdown_domains([self.qube], self.logger)
-            else:
-                self.logger.info('Shutdown %s', self.qube.name)
-                self.qube.shutdown()
+                wait = True
+            asyncio.run(shutdown_domains([self.qube], self.logger, wait=wait))
```
marmarek: Simply `self.qube.shutdown(wait=wait)` here, since it has `wait` option now?

marmarek (changes requested):
There is some conflict now.

vmupdate/qube_connection.py
ben-grande: I will push changes that resolve the merge conflict, and then will test if a freezed domain that tries to shutdown, will hang the progress of other qubes.

ben-grande: Testcase, log "SLEEP" and sleep before that part, `tail -F /var/log/qubes/update-QUBE.log`, start update for 3 domains, max concurrency of 2 and try to update. Freeze the qube when it's sleeping. Notice the update of the 3rd qube continue, so it's not freezing everything.

The difference is that that with my code, the exception message only appears in the logs, while with your proposal, it appears on the console. I think your proposal is correct. I think that a combination of these solutions are correct in this case, because it should be logged to both places and raised on the console.

marmarek:
now it has some conflicts (likely due to type hints)

PR #213, opened 2026-05-25
vmupdate: add qvm-template-upgrade

### LINKED ISSUE #8605: [Contribution]

Per official documentation, it is recommended that users download latest templates through `qvm-template` before "[switching](https://www.qubes-os.org/doc/templates/fedora/#upgrading)" everything from a pre-existing template to the newer template.  `qvm-prefs` allows for a somewhat seamless transition of [preferences](https://www.qubes-os.org/doc/templates/#switching). But inheritance and persistence is a bit more [complicated](https://www.qubes-os.org/doc/templates/#inheritance-and-persistence) for the non-technical user.

This makes upgrading-in-place a very attractive option for new and experienced users alike. [Fedrora documentation](https://www.qubes-os.org/doc/templates/fedora/in-place-upgrade/) and [Debian documentation](https://www.qubes-os.org/doc/templates/debian/in-place-upgrade/) suggest this process is for the "advanced" user. This should not be the case.

### The solution you'd like

Include as a qubes command the option to "upgrade-in-place" a template. I've provided basic Bash scripts for doing so, with an option for upgrading multiple templates of the same distro at once:

[For Fedora-based templates.](https://github.com/kennethrrosen/qubes-fedora-upgrader)
[For Debian-based templates.](https://github.com/kennethrrosen/qubes-debian-upgrader/tree/main)

At present, both scripts were tested successfully for upgrading Fedora-37 to Fedora-38 and Debian-11 to Debian-12. These scripts might also benefit from `qvm-clone` commands to create .bak templates. 

### The value to a user, and who that user might be
Such scripts would lower the barrier-to-entry for non-technical users interested in adopting Qubes. Furthermore, it would lower the amount of time spent on upgrading individual templates a user may have created based on the defaults, becoming a value-add.

nihalxkumar:
This PR introduces the `qvm-template-upgrade` dom0 command-line utility, that performs an in-place N -> N+1 distribution upgrade of Debian and Fedora TemplateVM or StandaloneVM

fixes: https://github.com/qubesOS/qubes-issues/issues/8605
GSoC 2026 project: [Automate Template Version Upgrade](https://summerofcode.withgoogle.com/programs/2026/projects/vjESxjOp)

### DIFF
--- a/doc/tools/qvm-template-upgrade.rst
+++ b/doc/tools/qvm-template-upgrade.rst
@@ -1,0 +1,73 @@
+====================
+qvm-template-upgrade
+====================
+
+NAME
+====
+qvm-template-upgrade - clone a qube and upgrade the clone to the next distribution release
+
+SYNOPSIS
+========
+| qvm-template-upgrade --template=<NAME> [OPTIONS]
+
+OPTIONS
+=======
+
+--template=<NAME>
+    Name of the source TemplateVM or StandaloneVM. This option is required.
+--new-name=<NAME>
+    Name for the clone. By default, replaces the last occurrence of the current ``os-version`` in the source name.
+--keep-new-on-failure
+    Keep the clone if the in-qube upgrade fails so that it can be inspected.
+--dry-run
+    Run dom0 preflight validation and print the planned source, clone name, distribution, and target release without creating a clone.
+--log=<LEVEL>
+    Set the workflow and in-qube agent log level.
+    Accepted values are ``DEBUG``, ``INFO`` (default), ``WARNING``, ``ERROR``, and ``CRITICAL``.
+-v, --verbose
+    Increase qubesadmin client verbosity.
+-q, --quiet
+    Decrease qubesadmin client verbosity.
+-h, --help
+    Show this help message and exit.
+
+CLONE NAME DERIVATION
+=====================
+
+Unless ``--new-name`` is specified, the last occurrence of the current version in the source name is replaced with the target version. If the current version is absent, the target version is appended:
+
+- ``fedora-41`` becomes ``fedora-42``.
+- ``fedora-41-minimal`` becomes ``fedora-42-minimal``.
+- ``custom-vm`` becomes ``custom-vm-42``.
+
+RETURN CODES
+============
+
+0:   The upgrade or dry run completed. A metadata-write warning after an otherwise successful upgrade also returns this status.
+
+1:   Clone creation, qrexec transport, or the in-qube upgrade failed.
+
+2:   Command-line parsing error.
+
+64:  Dom0 preflight validation error, such as an unsupported source qube, missing distribution features, a non-integer version, or an existing target name.
+
+130: Interrupted by the user. The clone may remain.
+
+FILES
+=====
+
+``/var/log/qubes/qvm-template-upgrade.log``
+    Main workflow log. If the file cannot be opened, the command continues with stderr logging only.
+
+``/var/log/qubes/update-<CLONE_NAME>.log``
+    Detailed qrexec transport and in-qube agent log. This file is created after the agent starts.
+
+SEE ALSO
+========
+
+**qubes-vm-update**\ (1), **qvm-clone**\ (1), **qvm-features**\ (1), **qvm-template**\ (1)
+
+AUTHORS
+=======
+| Nihal Kumar <nihalxkumar at tutamail dot com>
+| Ben Grande <ben at invisiblethingslab dot com>
--- a/doc/tools/qvm-template-upgrade.rst
+++ b/doc/tools/qvm-template-upgrade.rst
@@ -1,0 +1,102 @@
+====================
+qvm-template-upgrade
+====================
+
+NAME
+====
+qvm-template-upgrade - upgrade a clone to the next distribution release
+
+SYNOPSIS
+========
+| qvm-template-upgrade --template=<NAME> [OPTIONS]
+
+DESCRIPTION
+===========
+Clone the source qube and run the distribution release upgrade inside the
+clone. Agent output is streamed to the terminal above a progress bar and
+written to the log files listed in FILES. The progress bar is skipped when
+stderr is not a terminal.
+
+OPTIONS
+=======
+
+--template=<NAME>
+    Name of the source TemplateVM or StandaloneVM. This option is required.
+--new-name=<NAME>
+    Name for the clone. By default, replace the final ``os-version`` in the source name.
+--keep-new-on-failure
+    Keep the clone if the in-qube upgrade fails so that it can be inspected.
+--dry-run
+    Validate inputs and print the plan without creating a clone.
+--log=<LEVEL>
+    Set the workflow and in-qube agent log level.
+    Accepted values (case-insensitive) are ``DEBUG``, ``INFO`` (default), ``WARNING``, ``ERROR``, and ``CRITICAL``.
+-v, --verbose
+    Increase qubesadmin client verbosity.
+-q, --quiet
+    Decrease qubesadmin client verbosity.
+-h, --help
+    Show this help message and exit.
+
+CLONE NAME DERIVATION
+=====================
+
+Unless ``--new-name`` is set, replace the final current version. Append the
+target version when it is absent:
+
+- ``fedora-41`` becomes ``fedora-42``.
+- ``fedora-41-minimal`` becomes ``fedora-42-minimal``.
+- ``custom-vm`` becomes ``custom-vm-42``.
+
+TEMPLATE METADATA
+=================
+
+After a TemplateVM upgrade, update these ``template-*`` features:
+
+- ``template-name`` is set to the clone's name.
+- ``template-installtime`` and ``template-buildtime`` are set to the time of the upgrade, since the clone's root filesystem was produced then.
+- ``template-reponame`` is set to ``@qvm-template-upgrade``, marking the qube as neither installed from a repository nor from a local package.
+- ``template-summary`` and ``template-description`` have every occurrence of the source qube's name replaced with the clone's. Any other text is user-authored and is left alone.
+
+``template-epoch``, ``template-version``, and ``template-release`` stay
+inherited. **qvm-template**\ (1) uses them for package comparisons.
+
+StandaloneVMs are not managed by **qvm-template**\ (1), so none of their ``template-*`` features are modified.
+
+The inherited version makes **qvm-template**\ (1) list the clone as
+upgradeable. ``qvm-template upgrade`` would replace it with a stock template.
+
+RETURN CODES
+============
+
+0:   The upgrade or dry run completed. A metadata-write warning after an otherwise successful upgrade also returns this status.
+
+1:   Clone creation, qrexec transport, or the in-qube upgrade failed.
+
+2:   Command-line parsing error.
+
+64:  Dom0 preflight validation error, such as an unsupported source qube, a derivative distribution with its own version numbering, missing distribution features, a non-integer version, or an existing target name.
+
+130: Interrupted by the user. The clone may remain.
+
+FILES
+=====
+
+``/var/log/qubes/qvm-template-upgrade.log``
+    Main workflow log. If the file cannot be opened, the command continues with stderr logging only.
+
+``/var/log/qubes/update-<CLONE_NAME>.log``
+    Detailed qrexec transport and in-qube agent log. This file is created after the agent starts.
+
+SEE ALSO
+========
+
+**qubes-vm-update**\ (1), **qvm-clone**\ (1), **qvm-features**\ (1), **qvm-template**\ (1)
+
+AUTHORS
+=======
+This command was written by Nihal Kumar <nihalxkumar at tutamail dot com> as part of Google Summer of Code.
+
+Ben Grande <ben at invisiblethingslab dot com> reviewed the change.
+
+It was inspired by a personal project by Kenneth R. Rosen <kennethrrosen at proton dot me>.
--- a/vmupdate/agent/entrypoint.py
+++ b/vmupdate/agent/entrypoint.py
@@ -32,15 +32,35 @@ def main(args=None):
         os_data, log, log_handler, log_level, agent_type, args.no_progress
     )
 
-    log.debug("Running upgrades.")
-    return_code = pkg_mng.upgrade(
-        refresh=not args.no_refresh,
-        hard_fail=not args.force_upgrade,
-        remove_obsolete=not args.leave_obsolete,
-        print_streams=args.show_output,
-    )
+    if args.version_upgrade:
+        log.debug(
+            "Running distribution version upgrade to %s.",
+            args.version_upgrade,
+        )
+        try:
+            return_code = pkg_mng.version_upgrade(
+                args.version_upgrade, print_streams=args.show_output
+            )
+        except NotImplementedError as err:
+            log.error("Distribution version upgrade failed: %s", err)
+            print(str(err), file=sys.stderr)
+            return_code = EXIT.ERR_VM_UPDATE
+    else:
+        log.debug("Running upgrades.")
+        return_code = pkg_mng.upgrade(
+            refresh=not args.no_refresh,
+            hard_fail=not args.force_upgrade,
+            remove_obsolete=not args.leave_obsolete,
+            print_streams=args.show_output,
+        )
 
-    if not pkg_mng.PROGRESS_REPORTING and not args.no_progress:
+    # A version upgrade emits its own 0/100 milestones, so only the normal
+    # update path needs the fallback "finished" tick for CLI managers.
+    if (
+        not args.version_upgrade
+        and not pkg_mng.PROGRESS_REPORTING
+        and not args.no_progress
+    ):
         # even if progress reporting is unavailable we want info that update finished
         if agent_type is AgentType.UPDATE_VM:
             print(f"{55:.2f}", flush=True, file=sys.stderr)
--- a/vmupdate/agent/entrypoint.py
+++ b/vmupdate/agent/entrypoint.py
@@ -40,15 +40,35 @@ def main(args: list[str] | None = None) -> int:
         parsed_args.no_progress,
     )
 
-    log.debug("Running upgrades.")
-    return_code = pkg_mng.upgrade(
-        refresh=not parsed_args.no_refresh,
-        hard_fail=not parsed_args.force_upgrade,
-        remove_obsolete=not parsed_args.leave_obsolete,
-        print_streams=parsed_args.show_output,
-    )
+    if parsed_args.version_upgrade:
+        log.debug(
+            "Running distribution version upgrade to %s.",
+            parsed_args.version_upgrade,
+        )
+        try:
+            return_code = pkg_mng.version_upgrade(
+                parsed_args.version_upgrade,
+                print_streams=parsed_args.show_output,
+            )
+        except NotImplementedError as err:
+            log.error("Distribution version upgrade failed: %s", err)
+            print(str(err), file=sys.stderr)
+            return_code = EXIT.ERR_VM_UPDATE
+    else:
+        log.debug("Running upgrades.")
+        return_code = pkg_mng.upgrade(
+            refresh=not parsed_args.no_refresh,
+            hard_fail=not parsed_args.force_upgrade,
+            remove_obsolete=not parsed_args.leave_obsolete,
+            print_streams=parsed_args.show_output,
+        )
 
-    if not pkg_mng.PROGRESS_REPORTING and not parsed_args.no_progress:
+    # Version upgrades handle their own milestones; this is for normal updates only.
+    if (
+        not parsed_args.version_upgrade
+        and not pkg_mng.PROGRESS_REPORTING
+        and not parsed_args.no_progress
+    ):
         # even if progress reporting is unavailable we want info that update finished
         if agent_type is AgentType.UPDATE_VM:
             print(f"{55:.2f}", flush=True, file=sys.stderr)
--- a/vmupdate/agent/source/apt/apt_cli.py
+++ b/vmupdate/agent/source/apt/apt_cli.py
@@ -21,11 +21,12 @@
 
 # pylint: disable=unused-argument
 
+import csv
 import fcntl
 import glob
 import os
 import contextlib
-from typing import List
+from typing import List, Optional
 
 from source.utils import get_os_data
 from source.common.package_manager import PackageManager, AgentType
@@ -36,14 +37,9 @@ from source.common.exit_codes import EXIT
 class APTCLI(PackageManager):
     PROGRESS_REPORTING = False
 
-    # apt addresses releases by codename, but the tool speaks release
-    # numbers; extend as new releases appear
-    DEBIAN_CODENAMES = {
-        11: "bullseye",
-        12: "bookworm",
-        13: "trixie",
-        14: "forky",
-    }
+    # Maintained by Debian's distro-info-data package, which is a
+    # qubes-core-agent dependency (including in minimal templates).
+    DEBIAN_RELEASES_FILE = "/usr/share/distro-info/debian.csv"
 
     # Every apt source definition that can name a release codename, in both
     # the one-line (`.list`) and deb822 (`.sources`) formats.
@@ -210,7 +206,18 @@ class APTCLI(PackageManager):
             return guard
 
         old_codename = os_data["codename"]
-        new_codename = self.DEBIAN_CODENAMES[int(target)]
+        try:
+            new_codename = self._debian_codename(target)
+        except (OSError, csv.Error) as exc:
+            return self._refuse(
+                f"cannot read Debian release data from "
+                f"{self.DEBIAN_RELEASES_FILE}: {exc}."
+            )
+        if not new_codename:
+            return self._refuse(
+                f"no Debian codename for release {target!r} in "
+                f"{self.DEBIAN_RELEASES_FILE}."
+            )
 
         self._report_progress(0.0)
         result = ProcessResult()
@@ -280,14 +287,8 @@ class APTCLI(PackageManager):
         self, target: str, os_data: dict
     ) -> ProcessResult:
         """
-        The shared single-step check plus Debian specifics: the in-qube
-        family must be Debian and both codenames must be known.
+        The shared single-step check plus the current Debian codename check.
         """
-        if os_data.get("os_family") != "Debian":
-            return self._refuse(
-                f"in-qube OS family {os_data.get('os_family')!r} is not Debian."
-            )
-
         result = super()._verify_release_upgrade(target, os_data)
         if result.code:
             return result
@@ -296,13 +297,22 @@ class APTCLI(PackageManager):
             return self._refuse(
                 "cannot read the in-qube release codename from os-release."
             )
-        if int(target) not in self.DEBIAN_CODENAMES:
-            return self._refuse(
-                f"no known Debian codename for release {target!r}; "
-                "extend DEBIAN_CODENAMES in apt_cli.py."
-            )
 
         return ProcessResult()
+
+    @classmethod
+    def _debian_codename(cls, target: str) -> Optional[str]:
+        """Return Debian's codename for a numeric release."""
+        with open(
+            cls.DEBIAN_RELEASES_FILE,
+            "r",
+            encoding="utf-8",
+            newline="",
+        ) as releases_file:
+            for release in csv.DictReader(releases_file):
+                if release.get("version") == target:
+                    return release.get("series") or None
+        return None
 
     def _rewrite_sources(
         self, old_codename: str, new_codename: str
@@ -351,7 +361,7 @@ class APTCLI(PackageManager):
             return ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
         if not changed:
             return self._refuse(
-                f"no apt source references codename {old_codename!r}; refusing "
-                "to report a release upgrade that would not actually happen."
+                f"no apt source references codename {old_codename!r}; "
+                "refusing upgrade."
             )
         return ProcessResult()
--- a/vmupdate/agent/source/apt/apt_cli.py
+++ b/vmupdate/agent/source/apt/apt_cli.py
@@ -22,10 +22,12 @@
 # pylint: disable=unused-argument
 
 import fcntl
+import glob
 import os
 import contextlib
 from typing import List
 
+from source.utils import get_os_data
 from source.common.package_manager import PackageManager, AgentType
 from source.common.process_result import ProcessResult
 from source.common.exit_codes import EXIT
@@ -33,6 +35,23 @@ from source.common.exit_codes import EXIT
 
 class APTCLI(PackageManager):
     PROGRESS_REPORTING = False
+
+    # apt addresses releases by codename, but the tool speaks release
+    # numbers; extend as new releases appear
+    DEBIAN_CODENAMES = {
+        11: "bullseye",
+        12: "bookworm",
+        13: "trixie",
+        14: "forky",
+    }
+
+    # Every apt source definition that can name a release codename, in both
+    # the one-line (`.list`) and deb822 (`.sources`) formats.
+    APT_SOURCE_GLOBS = (
+        "/etc/apt/sources.list",
+        "/etc/apt/sources.list.d/*.list",
+        "/etc/apt/sources.list.d/*.sources",
+    )
 
     def __init__(self, log_handler, log_level, agent_type: AgentType):
         super().__init__(log_handler, log_level, agent_type)
@@ -173,3 +192,166 @@ class APTCLI(PackageManager):
 
         result.code = EXIT.ERR_VM_CLEANUP if result.code != 0 else 0
         return result
+
+    def _release_upgrade(self, target_version: str) -> ProcessResult:
+        """
+        Move the qube to the next Debian release.
+
+        apt addresses releases by codename (unlike dnf's ``--releasever``),
+        so after bringing the current release fully up to date we rewrite
+        the codename across all apt sources, then ``update`` +
+        ``dist-upgrade`` onto the new release.
+        """
+        target = str(target_version).strip()
+        os_data = get_os_data(self.log)
+
+        guard = self._verify_release_upgrade(target, os_data)
+        if guard.code:
+            return guard
+
+        old_codename = os_data["codename"]
+        new_codename = self.DEBIAN_CODENAMES[int(target)]
+
+        self._report_progress(0.0)
+        result = ProcessResult()
+
+        # bring the current release fully up to date, point the sources at
+        # the new release, then update + dist-upgrade onto it; a stale
+        # system risks unresolvable transactions across the boundary
+        steps = (
+            lambda: self.refresh(hard_fail=True),
+            lambda: self.upgrade_internal(remove_obsolete=False),
+            lambda: self._rewrite_sources(old_codename, new_codename),
+            lambda: self.refresh(hard_fail=True),
+            lambda: self.upgrade_internal(remove_obsolete=False),
+            self._dist_upgrade,
+        )
+        for step in steps:
+            step_result = step()
+            result += step_result
+            if step_result.code:
+                result.code = EXIT.ERR_VM_UPDATE
+                return result
+
+        try:
+            upgraded_os_data = get_os_data(self.log)
+        except OSError as exc:
+            msg = f"failed to verify the upgraded Debian release: {exc}"
+            self.log.error(msg)
+            result += ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
+            return result
+
+        upgraded_release = upgraded_os_data.get("release", "").split(".")[0]
+        upgraded_codename = upgraded_os_data.get("codename", "")
+        if upgraded_release != target or upgraded_codename != new_codename:
+            msg = (
+                f"Debian release upgrade did not reach {target} "
+                f"({new_codename}); os-release reports "
+                f"{upgraded_os_data.get('release')!r} "
+                f"({upgraded_codename or 'no codename'})."
+            )
+            self.log.error(msg)
+            result += ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
+            return result
+
+        # old kernels are heavy but non-critical: a cleanup failure must not
+        # discard a template that actually reached the new release
+        cleanup = self.remove_obsolete_kernels()
+        result += cleanup
+        if cleanup.code:
+            self.log.warning(
+                "post-upgrade obsolete-kernel cleanup failed (non-fatal): %s",
+                cleanup.err,
+            )
+        result.code = EXIT.OK
+
+        self._report_progress(100.0)
+        return result
+
+    def _dist_upgrade(self) -> ProcessResult:
+        """
+        Plain `apt-get dist-upgrade`, without APTCLI's obsolete-kernel
+        cleanup, so the fatal release bump and the best-effort cleanup
+        can fail independently.
+        """
+        return PackageManager.upgrade_internal(self, remove_obsolete=True)
+
+    def _verify_release_upgrade(
+        self, target: str, os_data: dict
+    ) -> ProcessResult:
+        """
+        The shared single-step check plus Debian specifics: the in-qube
+        family must be Debian and both codenames must be known.
+        """
+        if os_data.get("os_family") != "Debian":
+            return self._refuse(
+                f"in-qube OS family {os_data.get('os_family')!r} is not Debian."
+            )
+
+        result = super()._verify_release_upgrade(target, os_data)
+        if result.code:
+            return result
+
+        if not os_data.get("codename"):
+            return self._refuse(
+                "cannot read the in-qube release codename from os-release."
+            )
+        if int(target) not in self.DEBIAN_CODENAMES:
+            return self._refuse(
+                f"no known Debian codename for release {target!r}; "
+                "extend DEBIAN_CODENAMES in apt_cli.py."
+            )
+
+        return ProcessResult()
+
+    def _rewrite_sources(
+        self, old_codename: str, new_codename: str
+    ) -> ProcessResult:
+        """
+        Substring-replace the old codename with the new across every apt
+        source file, covering ``-security``/``-updates``/``-backports``
+        variants in both the one-line and deb822 formats. Missing files are
+        skipped; I/O errors abort.
+
+        Refuses if no file mentioned the old codename: such sources address
+        the release symbolically (e.g. ``stable``), the rewrite would be a
+        no-op and the qube would silently stay on the old release.
+
+        The plain substring replace matches the manual ``sed`` procedure in
+        qubes-doc; switch to per-field deb822 parsing if it ever over-matches.
+        """
+        changed = 0
+        try:
+            for pattern in self.APT_SOURCE_GLOBS:
+                for path in glob.glob(pattern):
+                    with open(path, "r", encoding="utf-8") as f_src:
+                        content = f_src.read()
+                    updated = content.replace(old_codename, new_codename)
+                    if updated == content:
+                        continue  # no codename here -> no spurious write
+                    # write-then-rename so a mid-write kill cannot leave a
+                    # truncated sources file behind (apt would treat an
+                    # empty sources list as "nothing to do" and succeed)
+                    tmp_path = path + ".tmp"
+                    with open(tmp_path, "w", encoding="utf-8") as f_src:
+                        f_src.write(updated)
+                        f_src.flush()
+                        os.fsync(f_src.fileno())
+                    os.replace(tmp_path, path)
+                    changed += 1
+                    self.log.info(
+                        "Rewrote apt source %s: %s -> %s",
+                        path,
+                        old_codename,
+                        new_codename,
+                    )
+        except OSError as exc:
+            msg = f"failed rewriting apt sources: {exc}"
+            self.log.error(msg)
+            return ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
+        if not changed:
+            return self._refuse(
+                f"no apt source references codename {old_codename!r}; refusing "
+                "to report a release upgrade that would not actually happen."
+            )
+        return ProcessResult()
--- a/vmupdate/agent/source/apt/apt_cli.py
+++ b/vmupdate/agent/source/apt/apt_cli.py
@@ -21,12 +21,15 @@
 
 # pylint: disable=unused-argument
 
+import csv
 import fcntl
+import glob
 import os
 import contextlib
-from typing import List, Iterator
+from typing import List, Iterator, Optional
 from logging import Handler
 
+from source.utils import get_os_data
 from source.common.package_manager import PackageManager, AgentType
 from source.common.process_result import ProcessResult
 from source.common.exit_codes import EXIT
@@ -34,6 +37,17 @@ from source.common.exit_codes import EXIT
 
 class APTCLI(PackageManager):
     PROGRESS_REPORTING = False
+
+    # Maintained by Debian's distro-info-data package, which is a
+    # qubes-core-agent dependency (including in minimal templates).
+    DEBIAN_RELEASES_FILE = "/usr/share/distro-info/debian.csv"
+
+    # APT source files in one-line (.list) and deb822 (.sources) formats.
+    APT_SOURCE_GLOBS = (
+        "/etc/apt/sources.list",
+        "/etc/apt/sources.list.d/*.list",
+        "/etc/apt/sources.list.d/*.sources",
+    )
 
     def __init__(
         self, log_handler: Handler, log_level: int, agent_type: AgentType
@@ -176,3 +190,192 @@ class APTCLI(PackageManager):
 
         result.code = EXIT.ERR_VM_CLEANUP if result.code != 0 else 0
         return result
+
+    def _release_upgrade(self, target_version: str) -> ProcessResult:
+        """Upgrade to the target Debian release by rewriting apt source codenames and running dist-upgrade."""
+        target = str(target_version).strip()
+        os_data = get_os_data(self.log)
+
+        guard = self._verify_release_upgrade(target, os_data)
+        if guard.code:
+            return guard
+
+        old_codename = os_data["codename"]
+        try:
+            new_codename = self._debian_codename(target)
+        except (OSError, csv.Error) as exc:
+            return self._refuse(
+                f"cannot read Debian release data from "
+                f"{self.DEBIAN_RELEASES_FILE}: {exc}."
+            )
+        if not new_codename:
+            return self._refuse(
+                f"no Debian codename for release {target!r} in "
+                f"{self.DEBIAN_RELEASES_FILE}."
+            )
+
+        # Refuse early if no source file references the old or new codename.
+        try:
+            sources = [
+                open(path, "r", encoding="utf-8").read()
+                for path in self._apt_source_paths()
+            ]
+        except OSError as exc:
+            return self._refuse(f"cannot read apt sources: {exc}.")
+        if not any(
+            old_codename in content or new_codename in content
+            for content in sources
+        ):
+            return self._refuse(
+                f"no apt source references codename {old_codename!r} "
+                f"or {new_codename!r}; refusing upgrade."
+            )
+
+        self._report_progress(0.0)
+        result = ProcessResult()
+
+        # Update before switching sources and running dist-upgrade: a stale
+        # system risks unresolvable transactions across the release boundary.
+        steps = (
+            lambda: self.refresh(hard_fail=True),
+            lambda: self.upgrade_internal(remove_obsolete=False),
+            lambda: self._rewrite_sources(old_codename, new_codename),
+            lambda: self.refresh(hard_fail=True),
+            lambda: self.upgrade_internal(remove_obsolete=False),
+            self._dist_upgrade,
+        )
+        for step in steps:
+            step_result = step()
+            result += step_result
+            if step_result.code:
+                result.code = EXIT.ERR_VM_UPDATE
+                return result
+
+        try:
+            upgraded_os_data = get_os_data(self.log)
+        except OSError as exc:
+            msg = f"failed to verify the upgraded Debian release: {exc}"
+            self.log.error(msg)
+            result += ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
+            return result
+
+        upgraded_release = upgraded_os_data.get("release", "").split(".")[0]
+        upgraded_codename = upgraded_os_data.get("codename", "")
+        if upgraded_release != target or upgraded_codename != new_codename:
+            msg = (
+                f"Debian release upgrade did not reach {target} "
+                f"({new_codename}); os-release reports "
+                f"{upgraded_os_data.get('release')!r} "
+                f"({upgraded_codename or 'no codename'})."
+            )
+            self.log.error(msg)
+            result += ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
+            return result
+
+        # Kernel cleanup failure must not discard a completed upgrade.
+        cleanup = self.remove_obsolete_kernels()
+        result += cleanup
+        if cleanup.code:
+            self.log.warning(
+                "post-upgrade obsolete-kernel cleanup failed (non-fatal): %s",
+                cleanup.err,
+            )
+        result.code = EXIT.OK
+
+        self._finish_progress()
+        return result
+
+    def _dist_upgrade(self) -> ProcessResult:
+        """Run dist-upgrade without kernel cleanup so failures are independent."""
+        return PackageManager.upgrade_internal(self, remove_obsolete=True)
+
+    def _verify_release_upgrade(
+        self, target: str, os_data: dict
+    ) -> ProcessResult:
+        """
+        The shared single-step check plus the current Debian codename check.
+        """
+        result = super()._verify_release_upgrade(target, os_data)
+        if result.code:
+            return result
+
+        if not os_data.get("codename"):
+            return self._refuse(
+                "cannot read the in-qube release codename from os-release."
+            )
+
+        return ProcessResult()
+
+    @classmethod
+    def _debian_codename(cls, target: str) -> Optional[str]:
+        """Return Debian's codename for a numeric release."""
+        with open(
+            cls.DEBIAN_RELEASES_FILE,
+            "r",
+            encoding="utf-8",
+            newline="",
+        ) as releases_file:
+            for release in csv.DictReader(releases_file):
+                if release.get("version") == target:
+                    if not release.get("release"):
+                        # distro-info lists unreleased versions too;
+                        # not a released Debian yet, so no codename.
+                        return None
+                    return release.get("series") or None
+        return None
+
+    @classmethod
+    def _apt_source_paths(cls) -> List[str]:
+        """Every existing apt source file matched by APT_SOURCE_GLOBS."""
+        return [
+            path
+            for pattern in cls.APT_SOURCE_GLOBS
+            for path in glob.glob(pattern)
+        ]
+
+    def _rewrite_sources(
+        self, old_codename: str, new_codename: str
+    ) -> ProcessResult:
+        """Replace the old release codename with the new one in all apt source files.
+
+        Refuses if no source references either codename (e.g. symbolic 'stable' sources).
+        """
+        changed = 0
+        try:
+            for path in self._apt_source_paths():
+                with open(path, "r", encoding="utf-8") as f_src:
+                    content = f_src.read()
+                updated = content.replace(old_codename, new_codename)
+                if updated == content:
+                    if new_codename in content:
+                        # Already rewritten by an interrupted run.
+                        changed += 1
+                    continue  # no codename here -> no spurious write
+                # Atomic write: preserve symlinks, permissions, and ownership.
+                real_path = os.path.realpath(path)
+                st = os.stat(real_path)
+                tmp_path = real_path + ".tmp"
+                with open(tmp_path, "w", encoding="utf-8") as f_src:
+                    f_src.write(updated)
+                    f_src.flush()
+                    os.fsync(f_src.fileno())
+                os.chmod(tmp_path, st.st_mode)
+                os.chown(tmp_path, st.st_uid, st.st_gid)
+                os.replace(tmp_path, real_path)
+                changed += 1
+                self.log.info(
+                    "Rewrote apt source %s: %s -> %s",
+                    path,
+                    old_codename,
+                    new_codename,
+                )
+        except OSError as exc:
+            msg = f"failed rewriting apt sources: {exc}"
+            self.log.error(msg)
+            return ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
+        if not changed:
+            return self._refuse(
+                f"no apt source references codename {old_codename!r}; "
+                "refusing upgrade."
+            )
+        return ProcessResult()
--- a/vmupdate/agent/source/apt/apt_cli.py
+++ b/vmupdate/agent/source/apt/apt_cli.py
@@ -240,15 +240,31 @@ class APTCLI(PackageManager):
 
         # Update before switching sources and running dist-upgrade: a stale
         # system risks unresolvable transactions across the release boundary.
+        # Each step gets its own slice of the bar, weighted by its share of
+        # a measured debian-12 to 13 template upgrade; without that, every
+        # step drives the same three-phase reporter to 100 in turn.
+        # (bar weight, drives fetch/install phases, step)
         steps = (
-            lambda: self.refresh(hard_fail=True),
-            lambda: self.upgrade_internal(remove_obsolete=False),
-            lambda: self._rewrite_sources(old_codename, new_codename),
-            lambda: self.refresh(hard_fail=True),
-            lambda: self.upgrade_internal(remove_obsolete=False),
-            self._dist_upgrade,
+            (7, False, lambda: self.refresh(hard_fail=True)),
+            (7, True, lambda: self.upgrade_internal(remove_obsolete=False)),
+            (
+                1,
+                False,
+                lambda: self._rewrite_sources(old_codename, new_codename),
+            ),
+            (9, False, lambda: self.refresh(hard_fail=True)),
+            (22, True, lambda: self.upgrade_internal(remove_obsolete=False)),
+            (54, True, self._dist_upgrade),
         )
-        for step in steps:
+        total = sum(weight for weight, _, _ in steps)
+        done = 0
+        for weight, installs, step in steps:
+            self._set_progress_step(
+                done / total * RELEASE_UPGRADE_ALMOST_DONE,
+                (done + weight) / total * RELEASE_UPGRADE_ALMOST_DONE,
+                installs,
+            )
+            done += weight
             step_result = step()
             result += step_result
             if step_result.code:
--- a/vmupdate/agent/source/apt/apt_cli.py
+++ b/vmupdate/agent/source/apt/apt_cli.py
@@ -30,7 +30,11 @@ from typing import List, Iterator, Optional
 from logging import Handler
 
 from source.utils import get_os_data
-from source.common.package_manager import PackageManager, AgentType
+from source.common.package_manager import (
+    PackageManager,
+    AgentType,
+    RELEASE_UPGRADE_ALMOST_DONE,
+)
 from source.common.process_result import ProcessResult
 from source.common.exit_codes import EXIT
 
@@ -282,7 +286,8 @@ class APTCLI(PackageManager):
             )
         result.code = EXIT.OK
 
-        self._finish_progress()
+        # version_upgrade() owns the jump to 100, after qubes.PostInstall.
+        self._report_milestone(RELEASE_UPGRADE_ALMOST_DONE)
         return result
 
     def _dist_upgrade(self) -> ProcessResult:
--- a/vmupdate/agent/source/common/package_manager.py
+++ b/vmupdate/agent/source/common/package_manager.py
@@ -78,6 +78,33 @@ class PackageManager:
                 print(result.err, file=sys.stderr, flush=True)
         return result.code
 
+    def version_upgrade(
+        self,
+        target_version: str,
+        print_streams: bool = False,
+    ) -> int:
+        """
+        Upgrade the distribution to the next major release.
+
+        Separate from `upgrade`, which only refreshes packages within the
+        current release. The family-specific work lives in
+        `_release_upgrade`.
+
+        :param target_version: the major release to move to, e.g. "42"
+        :param print_streams: dump captured output to std streams
+        :return: return code (0 on success)
+        """
+        result = self._release_upgrade(target_version)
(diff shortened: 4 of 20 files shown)


vmupdate/tests/test_template_upgrade.py
```diff
@@ -0,0 +1,157 @@
+# coding=utf-8
```
ben-grande, line 2: Python hashbang please.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class UpgradeError(Exception):
+    """Anything that prevents the upgrade from running or completing."""
+
+
+def parse_args(argv=None):
+    parser = argparse.ArgumentParser(
+        prog='qvm-template-upgrade',
+        description='Upgrade a TemplateVM to the next distro version.'
```
ben-grande: Or StandaloneVM. Update other places accordingly.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,242 @@
...
+    app = app or qubesadmin.Qubes()
+
+    try:
+        source_vm = validate_template(app, args.template)
+        distro, current = detect_distro(source_vm)
+        target = compute_target_version(current)
+        new_name = derive_clone_name(
+            source_vm.name, current, target, args.new_name)
+    except UpgradeError as err:
+        print(f"error: {err}", file=sys.stderr)
```
ben-grande: `parser.print_error()` from `QubesArgumentParser`.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def validate_template(app, name):
+    """Return the VM object if `name` is an existing TemplateVM, else raise."""
+    try:
+        vm = app.domains[name]
+    except KeyError:
+        raise UpgradeError(f"No such qube: {name}")
+    if vm.klass != 'TemplateVM':
+        raise UpgradeError(
```
ben-grande: I think that validation errors should raise a different exception than failed upgrade.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,242 @@
...
+
+def detect_distro(vm):
+    """Read distro family and current version from qvm-features."""
+    distro = vm.features.get('os-distribution')
+    distro_like = vm.features.get('os-distribution-like', '')
+    version = vm.features.get('os-version')
+    if not distro or not version:
+        raise UpgradeError(
+            f"{vm.name} is missing os-distribution / os-version features. "
+            f"Start the template once so the in-VM agent can report them, "
```
ben-grande: The `in-VM` agent probably should be aware of the distro also from inside the qube.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,242 @@
...
+    qvm-template EVR/buildtime features are intentionally preserved because
+    current qvm-template list/info paths require them for managed templates.
+    """
+    log.info("Updating template metadata on %s", clone_vm.name)
+    clone_vm.features['template-name'] = clone_vm.name
+    clone_vm.features['template-installtime'] = \
+        datetime.now(tz=timezone.utc).strftime(DATE_FMT)
+
+
+def remove_failed_clone(clone_vm, log):
```
ben-grande: You might want to include some functions as a method of a class, to avoid passing objects explicitly.

ben-grande (changes requested):
Yay, started. As I've done a review now, I will receive a notification every time you commit. Let me know when you need another look or have doubts by mentioning me.

vmupdate/tests/test_template_upgrade.py
nihalxkumar: Done. Thanks!

vmupdate/template_upgrade.py
nihalxkumar: Done in a0e15b7.
Added `SUPPORTED_CLASSES = {'TemplateVM', 'StandaloneVM'}`
`apply_post_upgrade_metadata` is a no-op for StandaloneVM since the `template-*` features only apply to managed TemplateVMs.

vmupdate/template_upgrade.py
nihalxkumar: Done. 
For the runtime failures (clone failure and agent failure after the validation phase) I kept plain `print(..., file=sys.stderr)` since they aren't usage errors

vmupdate/template_upgrade.py
nihalxkumar: added `ValidationError` for input/setup problems (mapped to `EXIT.ERR_USAGE`)
kept `UpgradeError` for failures during the upgrade run

vmupdate/template_upgrade.py
nihalxkumar: Agreed. I've noted this in the stub docstring (`run_upgrade_agent`) and will implement it as part of the agent extension in the next PR

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,260 @@
...
+    handler.setFormatter(logging.Formatter(LOG_FORMAT))
+    log.addHandler(handler)
+    # Also mirror to stderr so the user sees progress without tailing the log.
+    stderr = logging.StreamHandler(sys.stderr)
+    stderr.setFormatter(logging.Formatter('%(message)s'))
+    log.addHandler(stderr)
+    return log
+
+
+def validate_template(app, name):
```
ben-grande: Adapt the function name also to the new klass it encompasses.

nihalxkumar: done :saluting_face:

vmupdate/template_upgrade.py
nihalxkumar: done in https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/4cfd7e46efaa7695b173e635b07478c369fe8538

ben-grande:
PipelineRetryFailed

nihalxkumar:
It's showing a successful run

image

image

ben-grande:
Some tests are not enabled on this repo, such as mypy, black and pylint. See this as an example: https://github.com/QubesOS/qubes-core-admin/blob/main/.gitlab-ci.yml. Can you create a separate PR to enable those checks?

nihalxkumar:
Sure, will do

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,340 @@
...
+    """
+    if override:
+        return override
+    if current_version not in source_name:
+        raise ValidationError(
+            f"Cannot derive new template name from {source_name!r}: it does "
+            f"not contain the current version {current_version!r}. Pass "
+            f"--new-name explicitly.")
+    # Replace only the last occurrence (e.g. fedora-41-extras-41 stays sane).
+    head, _, tail = source_name.rpartition(current_version)
```
ben-grande, line 105: What if the template or standalone doesn't have the version in the name? Just append the version (`-$version`)?

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,340 @@
...
+    )
+    parser.add_argument(
+        '--template', required=True,
+        help='Name of the source TemplateVM or StandaloneVM to upgrade.')
+    parser.add_argument(
+        '--new-name',
+        help='Name for the upgraded clone. Defaults to replacing the version '
+             'suffix in the source name (e.g. fedora-41 -> fedora-42).')
+    parser.add_argument(
+        '--keep-on-failure', action='store_true',
```
ben-grande: `--keep-new-on-failure`, so the option name is explicit without having to read the description.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,340 @@
...
+        self.new_name = derive_clone_name(
+            self.source_vm.name,
+            self.current_version,
+            self.target_version,
+            self.args.new_name,
+        )
+        if self.new_name in self.app.domains:
+            raise ValidationError(
+                f"Target name {self.new_name!r} already exists. Remove it "
+                f"first or pass a different --new-name.")
```
ben-grande: When pylint is enabled, it will complain this f-string is unnecessary.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,340 @@
...
+                f"({self.distro} {self.current_version}) -> "
+                f"clone {self.new_name} "
+                f"({self.distro} {self.target_version})")
+
+    # execution
+
+    def clone(self):
+        """Clone the source qube. Populates self.clone_vm."""
+        self.log.info("Cloning %s -> %s", self.source_vm.name, self.new_name)
+        self.clone_vm = self.app.clone_vm(self.source_vm, self.new_name)
```
ben-grande: `self.clone_vm` is an action, used for functions. Use `self.cloned_qube` or `self.new_qube`.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,340 @@
...
+
+        StandaloneVM: rewrite an existing template-name from the old to
+        the new release (e.g. fedora-41 -> fedora-42), keeping the value
+        compatible with qui.utils.check_support()'s EOL_DATES lookup
+        (which strips only -minimal / -xfce, not -standalone or the qube
+        name). We do not invent a template-name for standalones that
+        never had one, and we leave one in place that doesn't carry the
+        current version (can't safely transform).
+
+        Inherited EVR/buildtime features are intentionally left in place
```
ben-grande: What is EVR?

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,340 @@
...
+        because qvm-template.query_local() uses bracket access on them and
```
ben-grande: To refer to code, refer to it how python would: `qubesadmin.tools.qvm_template.query_local`.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,340 @@
...
+        deleting would crash qvm-template list/info for this qube.
```
ben-grande: Yes, deleting them doesn't make sense, but them I think this decision does not need to be in the docstring.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,340 @@
...
+            vm = self.app.domains[self.args.template]
+        except KeyError:
+            raise ValidationError(f"No such qube: {self.args.template}")
+        if vm.klass not in SUPPORTED_CLASSES:
+            raise ValidationError(
+                f"{vm.name} is a {vm.klass}; only TemplateVMs and "
+                f"StandaloneVMs can be upgraded with this tool.")
+        return vm
+
+    def _detect_distro(self):
```
ben-grande: Please add typing hints to all functions.

ben-grande (changes requested):

nihalxkumar:
We can squash here if it's looks fine. Will also have to rebase as this is 24 commits behind.

ben-grande:
Looks fine to squash.

vmupdate/template_upgrade.py
nihalxkumar: Done as you suggested
custom -> custom-42

vmupdate/template_upgrade.py
nihalxkumar: Okay

vmupdate/template_upgrade.py
nihalxkumar: removed

vmupdate/template_upgrade.py
nihalxkumar: renamed to `self.cloned_qube`

vmupdate/template_upgrade.py
nihalxkumar: template metadata from `qvm-features` template-epoch, template-version, and template-release

image

vmupdate/template_upgrade.py
nihalxkumar: removed

vmupdate/template_upgrade.py
nihalxkumar: done, thanks!

marmarek:
Oh no, commit message in the other PR closed it ...

marmarek:
I see a conflict here - CI will not run until it's resolved.

vmupdate/agent/source/common/package_manager.py
```diff
@@ -78,6 +78,31 @@ def upgrade(
...
+
+        :param target_version: the major release to move to, e.g. "42"
+        :param print_streams: dump captured output to std streams
+        :return: return code (0 on success)
+        """
+        result = self._version_upgrade(target_version)
+        self._log_output("version-upgrade", result)
+        if print_streams and not result.posted:
+            if result.out:
+                print(result.out, flush=True)
```
ben-grande, line 103: Is there a reason some things are printed and others are logged?

vmupdate/agent/source/common/package_manager.py
```diff
@@ -312,6 +337,22 @@ def upgrade_internal(self, remove_obsolete: bool) -> ProcessResult:
...
+        families without a real implementation -- Debian/apt and Arch/pacman
+        today -- return a non-zero error here instead of silently reporting
+        success, so dom0 surfaces the failure and rolls back.
+        """
+        msg = (
+            "Distribution version upgrade is not implemented for this "
+            f"package manager ({self.package_manager})."
+        )
+        self.log.error(msg)
+        return ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
```
ben-grande, line 386: This function `_version_upgrade` can raise `NotImplementedError("Missing integration for ...")`, and the caller deals with the exception and loging if they want.

vmupdate/agent/source/dnf/dnf_cli.py
```diff
@@ -142,6 +144,114 @@ def get_action(self, remove_obsolete) -> List[str]:
                 result.append("update")
         return result
 
+    def _version_upgrade(self, target_version: str) -> ProcessResult:
```
ben-grande: Instead of naming two functions with the same name and prefixing one with `_`, I believe it's better to give explanatory names. Note there is a case for using `_` for internal functions, but I don't think it's the case here.

vmupdate/agent/source/dnf/dnf_cli.py
```diff
@@ -142,6 +144,114 @@ def get_action(self, remove_obsolete) -> List[str]:
...
+    def _refuse(self, reason: str) -> ProcessResult:
+        """Log and build the standard "refusing version upgrade" error."""
+        msg = f"Refusing version upgrade: {reason}"
+        self.log.error(msg)
+        return ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
+
+    @staticmethod
+    def _report_progress(percent: float) -> None:
+        """
+        Emit a progress milestone on the stderr-float channel that dom0's
```
ben-grande: What is "stderr-float channel"?

ben-grande (commented):
I have taken a look through the new code. Thanks for the progress, I like to see it evolving. Just minor review though

vmupdate/agent/source/common/package_manager.py
nihalxkumar: Existing pattern in `upgrade()` by Piotr

https://github.com/QubesOS/qubes-core-admin-linux/blob/497d467803249db2ef1e5a771f7e82a44bf941b2/vmupdate/agent/source/common/package_manager.py#L72-L81

per the vmupdate architecture, print writes to stdout/stderr which goes back to dom0 over qrexec (for the progress bar and live output) and logger writes to the VM-local log file for debugging. These are two separate channels.

the guard ( print_streams  +  not result.posted ) is just following the existing upgrade() pattern to avoid double-echoing when run_cmd already streamed in realtime.

vmupdate/agent/source/common/package_manager.py
nihalxkumar: Agreed.

vmupdate/agent/source/dnf/dnf_cli.py
nihalxkumar: I meant the qrexec stderr stream that `QubeConnection._collect_stderr` parses for progress floats. 
Updated the docstring for clarity.

vmupdate/agent/source/dnf/dnf_cli.py
nihalxkumar: replaced with `_release_upgrade` 
Open to change

ben-grande:
https://github.com/QubesOS/qubes-core-admin-linux/pull/213/changes/a0289a754c4936f215ac7f6920d678b4821ec213..a90eee15a3ce76eb66ca5d87eea4bf923789955e

Can you share more details of what fails on cleanup?

nihalxkumar:
When I tried upgrading fedora 41 -> 42 there was a cleanup failure which happened after the successful version upgrade.

logs as shared privately on tuesday:

https://gist.github.com/nihalxkumar/01fb990deab3960cb28d680773bb1089#file-upgrade42-log-L48-L55

we can see `Complete!` in the above log.

https://gist.github.com/nihalxkumar/0e4edccb4f4409b1a4daf5c44b6f576f

After `version_upgrade()` returns success, the agent still runs cleanup. The old code used the cleanup exit code when deciding the final agent exit code.
For dnf this includes `DNFCLI.clean()` (`dnf clean packages`), and the transport also removes `/run/qubes-update/`. 

[`a90eee1` (last commit)](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/a90eee15a3ce76eb66ca5d87eea4bf923789955e) is based on that failure boundary. After the release transaction has succeeded, a later cleanup failure should not cause rollback of the upgraded clone.

ben-grande:
But we can't ignore all cleanup failures. We don't know what is causing it, so this needs to be investigated.

nihalxkumar:
[a90eee1](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/a90eee15a3ce76eb66ca5d87eea4bf923789955e) wasn't ignoring all cleanup failures :sweat_smile: Anyways, we finally have a full end to end successful fedora upgrade pipeline

```
fedora-42-xfce:err: [3662/3664] Removing fedora-release-ide 100% |  21.0   B/s |   1.0   B |  00m00s
fedora-42-xfce:err: [3663/3664] Removing ncurses-base-0:6.5 100% |   2.6 KiB/s | 179.0   B |  00m00s
fedora-42-xfce:err: warning: posix.fork(): .fork(), .exec(), .wait() and .redirect2null() are deprecated, use rpm.spawn() or rpm.execute() instead
fedora-42-xfce:err: warning: posix.wait(): .fork(), .exec(), .wait() and .redirect2null() are deprecated, use rpm.spawn() or rpm.execute() instead
fedora-42-xfce:err: warning: posix.exec(): .fork(), .exec(), .wait() and .redirect2null() are deprecated, use rpm.spawn() or rpm.execute() instead
fedora-42-xfce:out: /etc/selinux/targeted/contexts/files/file_contexts.bin:  Old compiled fcontext format, skipping
fedora-42-xfce:out: /etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin:  Old compiled fcontext format, skipping
fedora-42-xfce:out: /etc/selinux/targeted/contexts/files/file_contexts.bin:  Old compiled fcontext format, skipping
fedora-42-xfce:out: /etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin:  Old compiled fcontext format, skipping
fedora-42-xfce:out: /usr/sbin cannot be merged yet, found /usr/sbin/capsh
fedora-42-xfce:out: All files under /usr/local/sbin are symlinks; linking to ./bin...
fedora-42-xfce:out: ...done
fedora-42-xfce:out: /usr/sbin cannot be merged yet, found /usr/sbin/capsh
fedora-42-xfce:err: [3664/3664] Removing libgcc-0:14.3.1-4. 100% |   0.0   B/s |  11.0   B |  25m26s
fedora-42-xfce:err: >>> Running %posttrans scriptlet: qubes-core-agent-selinux-0:4.2.47-1.fc42.noarc
fedora-42-xfce:err: >>> Finished %posttrans scriptlet: qubes-core-agent-selinux-0:4.2.47-1.fc42.noar
fedora-42-xfce:err: >>> [RPM] libselinux: type 0: /etc/selinux/targeted/contexts/files/file_contexts
fedora-42-xfce:err: >>> [RPM] libselinux: type 0: /etc/selinux/targeted/contexts/files/file_contexts
fedora-42-xfce:err: >>> Running %posttrans scriptlet: kernel-core-0:6.19.14-108.fc42.x86_64         
fedora-42-xfce:err: >>> Finished %posttrans scriptlet: kernel-core-0:6.19.14-108.fc42.x86_64        
fedora-42-xfce:err: >>> Scriptlet output:                                                           
fedora-42-xfce:err: >>> dracut[E]: Module 'systemd-cryptsetup' depends on module 'crypt', which can'
fedora-42-xfce:err: >>> dracut[E]: Module 'systemd-pcrphase' depends on module 'tpm2-tss', which can
fedora-42-xfce:err: >>>                                                                             
fedora-42-xfce:err: >>> Running %posttrans scriptlet: qubes-kernel-vm-support-0:4.2.22-1.fc42.x86_64
fedora-42-xfce:err: >>> Finished %posttrans scriptlet: qubes-kernel-vm-support-0:4.2.22-1.fc42.x86_6
fedora-42-xfce:err: >>> Scriptlet output:                                                           
fedora-42-xfce:err: >>> dracut[E]: Module 'systemd-cryptsetup' depends on module 'crypt', which can'
fedora-42-xfce:err: >>> dracut[E]: Module 'systemd-pcrphase' depends on module 'tpm2-tss', which can
fedora-42-xfce:err: >>> dracut[E]: Module 'systemd-cryptsetup' depends on module 'crypt', which can'
fedora-42-xfce:err: >>> dracut[E]: Module 'systemd-pcrphase' depends on module 'tpm2-tss', which can
fedora-42-xfce:err: >>> dracut[E]: Module 'systemd-cryptsetup' depends on module 'crypt', which can'
fedora-42-xfce:err: >>> dracut[E]: Module 'systemd-pcrphase' depends on module 'tpm2-tss', which can
fedora-42-xfce:err: >>>                                                                             
fedora-42-xfce:err: >>> Running %triggerin scriptlet: systemd-0:257.13-1.fc42.x86_64                
fedora-42-xfce:err: >>> Finished %triggerin scriptlet: systemd-0:257.13-1.fc42.x86_64               
fedora-42-xfce:err: >>> Scriptlet output:                                                           
fedora-42-xfce:err: >>>                                                                             
fedora-42-xfce:err: Complete!
Updating metadata on fedora-42-xfce
Upgrade complete. New template: fedora-42-xfce
Original qube fedora-41-xfce is untouched.
[nxk@dom0 ~]$ 
```

More logs are at https://gist.github.com/nihalxkumar/e02475c11b6dc8a2e855ceec66599d49

For the fix had to extend the try block in https://github.com/QubesOS/qubes-core-admin-linux/blob/497d467803249db2ef1e5a771f7e82a44bf941b2/vmupdate/update_manager.py#L362-L366

I have mentioned details in the [f66d1e7](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/f66d1e707f19629523b9de93fbe748792b32f5e2)'s description.

ben-grande:
> [a90eee1](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/a90eee15a3ce76eb66ca5d87eea4bf923789955e) wasn't ignoring all cleanup failures 😅

https://github.com/QubesOS/qubes-core-admin-linux/commit/a90eee15a3ce76eb66ca5d87eea4bf923789955e

How not? It was ignoring all cleanup failures if `pkg_mng.version_upgrade` returned `EXIT.OK`. Can you explain in more details? I didn't check if this change remained in the latest  commits.

ben-grande:
https://github.com/QubesOS/qubes-core-admin-linux/commit/f66d1e707f19629523b9de93fbe748792b32f5e2

Unsure about this one also. If shutdown failed, the upgrade might have failed despite the package manager not returning an error during upgrade. Which exception were you receiving?

nihalxkumar:
> > [a90eee1](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/a90eee15a3ce76eb66ca5d87eea4bf923789955e) wasn't ignoring all cleanup failures 😅
> 
> [a90eee1](https://github.com/QubesOS/qubes-core-admin-linux/commit/a90eee15a3ce76eb66ca5d87eea4bf923789955e)
> 
> How not? It was ignoring all cleanup failures if `pkg_mng.version_upgrade` returned `EXIT.OK`. Can you explain in more details? I didn't check if this change remained in the latest commits.

By "all" I thought we are including the regular-updates cleanup behavior as well, which were untouched. On the version-upgrade path, though, you're right: once the upgrade returned `EXIT.OK`, [a90eee1](https://github.com/QubesOS/qubes-core-admin-linux/commit/a90eee15a3ce76eb66ca5d87eea4bf923789955e) suppressed any non-zero `clean()` code. My reasoning was that a failing `clean()` after a completed release transaction is just a cache issue, but the dom0 tool rolls back on any agent exit `!= EXIT.OK` ([`template_upgrade.py#L307`](https://github.com/QubesOS/qubes-core-admin-linux/blob/bd9a7236a6e8c2ca0b4da9e5294cc069ac2eb4fa/vmupdate/template_upgrade.py#L307)) a fatal cleanup code would delete a good upgrade. So I tolerated it only on the success path and kept it fatal for normal updates.

I've since dropped that tolerance change. As in every run `dnf clean` actually returned 0, so it was never being triggered; and the rollbacks I was chasing were exit-26 (`ERR_VM_UNHANDLED`), whereas a cleanup failure can only ever be exit-25 (`ERR_VM_CLEANUP`) ([`dnf_cli.py#L145-L151`](https://github.com/QubesOS/qubes-core-admin-linux/blob/497d467803249db2ef1e5a771f7e82a44bf941b2/vmupdate/agent/source/dnf/dnf_cli.py#L145-L151)), which is already in `VM_HANDLED` ([`exit_codes.py#L34-L40`](https://github.com/QubesOS/qubes-core-admin-linux/blob/497d467803249db2ef1e5a771f7e82a44bf941b2/vmupdate/agent/source/common/exit_codes.py#L34-L40)) and never escalates to 26. So cleanup wasn't the cause. The real source of exit-26 was the teardown, which [`bd9a723`](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/bd9a7236a6e8c2ca0b4da9e5294cc069ac2eb4fa) addresses.

nihalxkumar:
> [f66d1e7](https://github.com/QubesOS/qubes-core-admin-linux/commit/f66d1e707f19629523b9de93fbe748792b32f5e2)
> 
> Unsure about this one also. If shutdown failed, the upgrade might have failed despite the package manager not returning an error during upgrade. Which exception were you receiving?

The exception was an `AttributeError`:

```
2026-06-26 07:52:44,585 Cannot shutdown fedora-42-xfce, because of error: 'DeviceCollection' object has no attribute 'get_assigned_devices'
```

It's raised in the pre-existing `_has_assigned_pci_devices` helper ([`qube_connection.py#L109-L112`](https://github.com/QubesOS/qubes-core-admin-linux/blob/497d467803249db2ef1e5a771f7e82a44bf941b2/vmupdate/qube_connection.py#L109-L112), added in [`12e9e32`](https://github.com/QubesOS/qubes-core-admin-linux/commit/12e9e326d99121394cc339ddc0cfbaae64ce082a) 3 months ago) while `__exit__` shutsdown the clone.

To your concern that a failed shutdown could mean a failed upgrade: in this case it didn't. The agent finishes cleanly: `distro-sync` exit 0, `dnf clean packages` exit 0 and the failure only appears afterwards, at shutdown ([logs shared earlier](https://gist.github.com/nihalxkumar/e02475c11b6dc8a2e855ceec66599d49#file-update-fedora-42-xfce-log)):


```
2026-06-26 05:40:02,642 [Agent] run command: dnf clean all
2026-06-26 05:40:04,387 [Agent] command exit code: 0
2026-06-26 05:40:04,391 [Agent] run command: dnf --releasever=42 distro-sync --best --allowerasing --assumeyes
2026-06-26 07:52:09,778 [Agent] command exit code: 0
2026-06-26 07:52:09,798 [Agent] version-upgrade out: Removed 24 files, 11 directories (total of 87 MiB). 0 errors occurred.
2026-06-26 07:52:09,808 [Agent] version-upgrade out: 
2026-06-26 07:52:09,817 [Agent] Notify dom0 about upgrades.
2026-06-26 07:52:38,827 [Agent] run command: dnf clean packages
2026-06-26 07:52:39,630 [Agent] command exit code: 0
2026-06-26 07:52:43,427 Remove /run/qubes-update/
2026-06-26 07:52:43,427 run command in fedora-42-xfce: rm -r /run/qubes-update/
2026-06-26 07:52:43,428 Wait for output
2026-06-26 07:52:44,585 Cannot shutdown fedora-42-xfce, because of error: 'DeviceCollection' object has no attribute 'get_assigned_devices'
2026-06-26 07:52:44,586 agent output: 
2026-06-26 07:52:44,586 agent output: 
2026-06-26 07:52:44,587 agent exit code: 0
```


My dom0 is on `qubes-core-admin-client-4.2.18`, whose [`DeviceCollection`](https://github.com/QubesOS/qubes-core-admin-client/blob/release4.2/qubesadmin/devices.py) still exposes `assignments()`/`attached()` and not `get_assigned_devices()`. The latter is the [R4.3](https://github.com/QubesOS/qubes-core-admin-client/blob/release4.3/qubesadmin/devices.py)/[`main`](https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/devices.py) device API that this branch targets, so on an actual R4.3 dom0 this specific `AttributeError` shouldn't fire.

[`bd9a723`](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/bd9a7236a6e8c2ca0b4da9e5294cc069ac2eb4fa) logs the shutdown error instead of letting a teardown exception propagate out of `_run_agent`, get caught by `update_qube`'s broad `except`, and surface as `ERR_VM_UNHANDLED` which then rolls back and deletes a clone whose upgrade already completed. It also appends the agent's `out`/`err` to the `UpgradeError`, so the real cause reaches the tool log next time instead of being discarded.

We could also harden the source instead, `_has_assigned_pci_devices` fall back across both device APIs (or catch `AttributeError` there) so the shutdown path itself is robust.

ben-grande:
> My dom0 is on `qubes-core-admin-client-4.2.18`, whose [`DeviceCollection`](https://github.com/QubesOS/qubes-core-admin-client/blob/release4.2/qubesadmin/devices.py) still exposes `assignments()`/`attached()` and not `get_assigned_devices()`. The latter is the [R4.3](https://github.com/QubesOS/qubes-core-admin-client/blob/release4.3/qubesadmin/devices.py)/[`main`](https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/devices.py) device API that this branch targets, so on an actual R4.3 dom0 this specific `AttributeError` shouldn't fire.

Can you upgrade to R4.3? It would make testing much easier, cause currently you are testing an EOL release.

> [`bd9a723`](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/commits/bd9a7236a6e8c2ca0b4da9e5294cc069ac2eb4fa) logs the shutdown error instead of letting a teardown exception propagate out of `_run_agent`, get caught by `update_qube`'s broad `except`, and surface as `ERR_VM_UNHANDLED` which then rolls back and deletes a clone whose upgrade already completed. It also appends the agent's `out`/`err` to the `UpgradeError`, so the real cause reaches the tool log next time instead of being discarded.

The exception should be raised so the command fails. It won't be raised when you have R4.3.

nihalxkumar:
I upgraded my system to R4.3 and the upgrade ran successfully end to end upgrade with no errors. I have reset the branch to d3ab852 as of june15.

The shutdown exception propagates, and failure detail it was trying to surface is already in `/var/log/qubes/update-<name>.log`, so no replacement was needed.

Next thing is disk handling for temporary cache volume

vmupdate/agent/source/dnf/dnf_cli.py
```diff
@@ -142,6 +144,117 @@ def get_action(self, remove_obsolete) -> List[str]:
...
+        be exactly one greater than the current in-qube major release. This
+        mirrors dom0's `compute_target_version`, ruling out downgrades,
+        no-ops, and multi-step jumps.
+        """
+        if not target.isdigit():
+            return self._refuse(f"invalid target release {target!r}.")
+
+        os_data = get_os_data(self.log)
+        family = os_data.get("os_family")
+        if family != "RedHat":
```
ben-grande: Did you find a case where this happens? We are in the Dnf module, Dnf was already chosen, and here we are checking again the `os_family` for which specific reason?

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,432 @@
...
+                f"{self.source_vm.name} is missing os-distribution / "
+                f"os-version features. Start the qube once so the in-VM "
+                f"agent can report them, then retry."
+            )
+        candidates = {distro.lower(), *distro_like.lower().split()}
+        supported = SUPPORTED_DISTROS & candidates
+        if not supported:
+            raise ValidationError(
+                f"Unsupported distro {distro!r}; only Fedora- and "
+                f"Debian-based qubes are supported for now."
```
ben-grande: Instead of writing the supported distros as strings, use something like this untested code: `[s.capitalize() for s in SUPPORTED_DISTROS]`

```python3
                f"Unsupported distro {distro!r}; Supported distros are ...
```

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,432 @@
...
+        """Remove the half-upgraded clone, if any. Safe to call repeatedly."""
+        if self.cloned_qube is None:
+            return
+        self.log.warning("Removing failed clone %s", self.cloned_qube.name)
+        try:
+            # The transport only requests an async shutdown, so the clone may
+            # still be running; Qubes refuses to delete a running VM, so force
+            # it down and wait before removing it.
+            if self.cloned_qube.is_running():
+                shutdown_domains([self.cloned_qube], self.log)
```
ben-grande: I think `kill()` is appropriate here as we don't care about this qube anymore.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,432 @@
...
+            )
+        print(f"error: {err}", file=sys.stderr)
+        return EXIT.ERR
+
+    label = (
+        "template"
+        if upgrader.cloned_qube.klass == "TemplateVM"
+        else "standalone"
+    )
+    print(f"Upgrade complete. New {label}: {upgrader.cloned_qube.name}")
```
ben-grande, line 440: label value should be `upgrader.cloned_qube.klass.lower()`, so even if ther eare other klass supported in the future, this would handle it.

ben-grande (changes requested):
Hi. Could you also take a look at:

- Debian integration
- API usage for better progress reporting

ben-grande:
In the future, this tool will be tested by openqa and when it is good enough, it should serve this use case:

- https://github.com/QubesOS/openqa-tests-qubesos/blob/152c563a4aea5eaa5446354f410fb5a4fa64af76/tests/distupgrade_template.pm#L41

In other words, it should be able to do, what this command from the link, does.

---

After you add API support, check if upgrading simultaneously with CLI and API breaks things or it is safely handled.

vmupdate/template_upgrade.py
nihalxkumar: Sure. I went with `klass.lower().removesuffix("vm")` so that output to user is  "New template:" / "New standalone:" instead of "New templatevm:" / "New standalonevm:"
You can take the final call

vmupdate/template_upgrade.py
nihalxkumar: Makes sense. While doing this I dropped the preceding `is_running()` guard too. `kill()` already raises `QubesVMNotStartedError` when the qube is halted, so the extra check was a redundant round-trip.

vmupdate/template_upgrade.py
nihalxkumar: Sure

vmupdate/template_upgrade.py
ben-grande: Your's is fine.

nihalxkumar:
@ben-grande Should debian integration be also included in this PR?

ben-grande:
Yes please.

nihalxkumar:
[`1271315` (fedora disk handling)](https://github.com/QubesOS/qubes-core-admin-linux/pull/213/changes/1271315d8b20e372ed0cdbc89f9072cdec093e71) automates the [manual](https://doc.qubes-os.org/en/latest/user/templates/fedora/fedora-upgrade.html#detailed-instructions-for-standard-fedora-templates) cache-volume procedure. I went with attaching a throwaway cache volume to the clone every time instead of just as a fallback when dnf complains about space because a whole-release distro-sync on a template's small root disk will basically always need it, so it's provisioned upfront instead of waiting for the first attempt to fail. 

dom0 creates a sparse 5GB image on disk, formats it ext4, and attaches it to the clone as a loop device through qvm-block, same as the manual steps. The one deliberate difference from the doc is that the agent mounts it by filesystem UUID rather than a device path like /dev/xvdi which isn't guaranteed stable.

Everything is torn down afterwards regardless of outcome: the loop device is detached and the image deleted on success, failure, or rollback, and a lock file lets a later run detect and reclaim an image left behind by one that got killed mid-way.

Open to design decisions, after the reviews I will test it on my system and then we can take on Debian upgrade.

vmupdate/agent/source/dnf/dnf_cli.py
```diff
@@ -204,6 +239,64 @@ def _release_upgrade(self, target_version: str) -> ProcessResult:
...
+        Wait up to `timeout` seconds for a device node to appear.
+
+        Attaching the volume in dom0 and its node showing up in the qube
+        race against each other, so poll rather than assume it is present.
+        """
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
+            if os.path.exists(path):
+                return True
+            time.sleep(interval)
```
ben-grande: Yeah, I don't like sleep but it seems the way to go, as used by the USB extension, until we have a better API.

- https://github.com/QubesOS/qubes-app-linux-usb-proxy/blob/207cf66afac157913cddd4896fb5c910d1a50e90/src/usb-export#L164
- https://github.com/QubesOS/qubes-app-linux-usb-proxy/blob/207cf66afac157913cddd4896fb5c910d1a50e90/src/usb-import#L86

vmupdate/agent/source/dnf/dnf_cli.py
```diff
@@ -204,6 +239,64 @@ def _release_upgrade(self, target_version: str) -> ProcessResult:
...
+        path: str, timeout: float = 10.0, interval: float = 0.5
+    ) -> bool:
+        """
+        Wait up to `timeout` seconds for a device node to appear.
+
+        Attaching the volume in dom0 and its node showing up in the qube
+        race against each other, so poll rather than assume it is present.
+        """
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
```
ben-grande: As it is a shorter loop, I think it is fine to keep getting the current time, but for more complex cases, store the amount of time you have slept and check it against timeout.

```python3
slept = 0
while slept < timeout:
    if os.path.exists(path):
        return True
    time.sleep(interval)
    slept += interval
return os.path.exists(path)
```

vmupdate/agent/entrypoint.py
```diff
@@ -39,7 +39,9 @@ def main(args=None):
         )
         try:
             return_code = pkg_mng.version_upgrade(
-                args.version_upgrade, print_streams=args.show_output
+                args.version_upgrade,
+                print_streams=args.show_output,
+                cache_device_uuid=args.dnf_cache_uuid,
```
ben-grande: I don't think this is something that users should pass on the command-line. I think there is a way to momentarily have extra space on a qube by creating a new mountpoint without requiring a dom0 device to be attached from dom0. Shared more on Matrix about a possible alternative.

vmupdate/agent/source/apt/apt_cli.py
```diff
@@ -173,3 +192,145 @@ def remove_obsolete_kernels(self) -> ProcessResult:
...
+        The plain substring replace matches the manual ``sed`` procedure in
+        qubes-doc; switch to per-field deb822 parsing if it ever over-matches.
+        """
+        changed = 0
+        try:
+            for pattern in self.APT_SOURCE_GLOBS:
+                for path in glob.glob(pattern):
+                    with open(path, "r", encoding="utf-8") as f_src:
+                        content = f_src.read()
+                    updated = content.replace(old_codename, new_codename)
```
ben-grande, line 339: Unfotunately, as far as I know, this is the way to replace deb repo defintions since ever. Have you searched if Debian supports a better way to do this?

vmupdate/agent/source/apt/apt_cli.py
```diff
@@ -173,3 +192,145 @@ def remove_obsolete_kernels(self) -> ProcessResult:
...
+                        new_codename,
+                    )
+        except OSError as exc:
+            msg = f"failed rewriting apt sources: {exc}"
+            self.log.error(msg)
+            return ProcessResult(EXIT.ERR_VM_UPDATE, out="", err=msg)
+        if not changed:
+            return self._refuse(
+                f"no apt source references codename {old_codename!r}; refusing "
+                "to report a release upgrade that would not actually happen."
```
ben-grande: Please improve this message. What do you think of, after the `;` just, ` refusing upgrade`.

vmupdate/agent/source/apt/apt_cli.py
```diff
@@ -173,3 +192,145 @@ def remove_obsolete_kernels(self) -> ProcessResult:
...
+    def _verify_release_upgrade(
+        self, target: str, os_data: dict
+    ) -> ProcessResult:
+        """
+        The shared single-step check plus Debian specifics: the in-qube
+        family must be Debian and both codenames must be known.
+        """
+        if os_data.get("os_family") != "Debian":
+            return self._refuse(
+                f"in-qube OS family {os_data.get('os_family')!r} is not Debian."
```
ben-grande: Same comment I made on Debina. We are already in the `apt` plugin, why check OS family?

vmupdate/agent/source/apt/apt_cli.py
```diff
@@ -173,3 +192,145 @@ def remove_obsolete_kernels(self) -> ProcessResult:
...
+            return result
+
+        if not os_data.get("codename"):
+            return self._refuse(
+                "cannot read the in-qube release codename from os-release."
+            )
+        if int(target) not in self.DEBIAN_CODENAMES:
+            return self._refuse(
+                f"no known Debian codename for release {target!r}; "
+                "extend DEBIAN_CODENAMES in apt_cli.py."
```
ben-grande: Does Debian maintain a file somewhere with a mapping of it's codenames so we don't have to keep updating this variable every one or two years?

ben-grande: See `/usr/share/distro-info/debian.csv`. Unfortunately it is a file from a package instead of being in base, package `distro-info-data`. Check if that package is installed by default on minimal templates.

Found another file, that seems to be in base, but please check it's presence on a minimal template: `/usr/lib/python3/dist-packages/debian/debian_support.py`. Also see if there is some goodies in that module that you could use.

ben-grande (changes requested):

doc/tools/qvm-template-upgrade.rst
```diff
@@ -0,0 +1,73 @@
+====================
```
parulin, line 1: Take a look at `doc/manpages/qvm-services.rst` for reStructuredText syntax example. `program` and `option` directives might be very useful for cross-referencing.

doc/tools/qvm-template-upgrade.rst
```diff
@@ -0,0 +1,73 @@
...
+qvm-template-upgrade
+====================
+
+NAME
+====
+qvm-template-upgrade - clone a qube and upgrade the clone to the next distribution release
+
+SYNOPSIS
+========
+| qvm-template-upgrade --template=<NAME> [OPTIONS]
```
parulin, line 11: Use something like:

```rst
:program:`qvm-template-upgrade` :option:`--template=<NAME>` ``[OPTIONS]``
```

doc/tools/qvm-template-upgrade.rst
```diff
@@ -0,0 +1,73 @@
...
+
+OPTIONS
+=======
+
+--template=<NAME>
```
parulin, line 16: Use `option` directive here and for the others if possible

doc/tools/qvm-template-upgrade.rst
```diff
@@ -0,0 +1,73 @@
+====================
+qvm-template-upgrade
+====================
+
+NAME
```
parulin, line 5: No need to capitalize?

doc/tools/qvm-template-upgrade.rst
```diff
@@ -0,0 +1,73 @@
...
+Unless ``--new-name`` is specified, the last occurrence of the current version in the source name is replaced with the target version. If the current version is absent, the target version is appended:
+
+- ``fedora-41`` becomes ``fedora-42``.
+- ``fedora-41-minimal`` becomes ``fedora-42-minimal``.
+- ``custom-vm`` becomes ``custom-vm-42``.
+
+RETURN CODES
+============
+
+0:   The upgrade or dry run completed. A metadata-write warning after an otherwise successful upgrade also returns this status.
```
parulin, line 46: A definition list like the one you used for options might be better here?

doc/tools/qvm-template-upgrade.rst
```diff
@@ -0,0 +1,73 @@
...
+
+130: Interrupted by the user. The clone may remain.
+
+FILES
+=====
+
+``/var/log/qubes/qvm-template-upgrade.log``
+    Main workflow log. If the file cannot be opened, the command continues with stderr logging only.
+
+``/var/log/qubes/update-<CLONE_NAME>.log``
```
parulin, line 62: Using:

```rst
:file:`/var/log/qubes/update-{<CLONE_NAME>}.log`
```

Allows you to emphasize the part to replace

doc/tools/qvm-template-upgrade.rst
```diff
@@ -0,0 +1,73 @@
...
+    Detailed qrexec transport and in-qube agent log. This file is created after the agent starts.
+
+SEE ALSO
+========
+
+**qubes-vm-update**\ (1), **qvm-clone**\ (1), **qvm-features**\ (1), **qvm-template**\ (1)
```
parulin, line 68: Is it possible to replace this by real references? like :doc:`qvm-template`?

parulin (commented):
I have some suggestion for better integration with the rest of the docs. If you have any question, feel free to ask.

This file is in `doc/tools` instead of `doc/manpages`, why?

parulin:
As I understand, this tool should also be promoted in qubes-doc, in the `user/templates/*/*-upgrade.rst` files for fedora and debian.

nihalxkumar:
@parulin you are mistaking this repository as qubes-core-admin-client

doc/tools/qvm-template-upgrade.rst
nihalxkumar: I followed the existing pattern in  https://github.com/QubesOS/qubes-core-admin-linux/blob/main/doc/tools/qubes-vm-update.rst
But yes, can do as a definition list.

@ben-grande thoughts?

parulin: Well, in that case it might be better to keep the pattern consistent.

ben-grande: Yes, keep consistent pattern. When there is no pattern, follow Parulin instructions on the remaining comments.

nihalxkumar: remaining comments are not relevant to our repository

parulin: Yes, sorry for the mess. As stated by @nihalxkumar I thought that we were in a sphinx documentation so it's not relevant at all!

nihalxkumar: No problem parulin. Thank you for the initiative 😄

nihalxkumar:
Debian upgrade end-to-end successful

https://gist.github.com/nihalxkumar/3f64e3a056e9ce367e9787d5bd94b9df
https://gist.github.com/nihalxkumar/212e885811589b85fdd23dafdfddf887

vmupdate/agent/entrypoint.py
nihalxkumar: As decided on matrix, the disk handling plan is being dropped. [Thread](https://matrix.to/#/!WtRrlYUTHOQjqGcSnn:invisiblethingslab.com/$UzGeAFpL7qkIZFqEKEA_zZbx3C1vWm6cOlqUAgejY_U?via=invisiblethingslab.com&via=matrix.org&via=nitro.chat)

vmupdate/agent/source/apt/apt_cli.py
nihalxkumar: Apparently this is it. Kept the substring replace and documented why in the `_rewrite_sources` docstring: it mirrors that sed workflow, and we can switch to per-field deb822 parsing if substring matching ever over-matches (say, a codename in a comment or URL). Also hardened the write with temp-file + fsync + os.replace so a kill mid-rewrite can't truncate a sources file.

vmupdate/agent/source/apt/apt_cli.py
nihalxkumar: Done

vmupdate/agent/source/apt/apt_cli.py
nihalxkumar: Yes, I was unsure but the upgrade did succeed after removing it.

vmupdate/agent/source/apt/apt_cli.py
nihalxkumar: Good find. We are now using debian.csv 
The package is installed by default as a dependency in minimal templates

doc/tools/qvm-template-upgrade.rst
nihalxkumar: Not relevant to this repository.
Thanks for the initiative

doc/tools/qvm-template-upgrade.rst
nihalxkumar: Not relevant to this repository.
Thanks for the initiative

doc/tools/qvm-template-upgrade.rst
nihalxkumar: Not relevant to this repository.
Thanks for the initiative

doc/tools/qvm-template-upgrade.rst
nihalxkumar: Not relevant to this repository.
Thanks for the imitative

doc/tools/qvm-template-upgrade.rst
nihalxkumar: Not relevant to this repository.
Thanks for the initiative

doc/tools/qvm-template-upgrade.rst
nihalxkumar: Not relevant to this repository.
Thanks for the initiative

doc/tools/qvm-template-upgrade.rst
```diff
@@ -0,0 +1,73 @@
...
+
+AUTHORS
+=======
+| Nihal Kumar <nihalxkumar at tutamail dot com>
+| Ben Grande <ben at invisiblethingslab dot com>
```
deeplow, line 73: Not sure how credit works here, but it could be worth also adding @kennethrrosen as the creator [of the tool](https://github.com/kennethrrosen/qubes-template-upgrade/blob/main/qvm-template-upgrade) that lead to this. 🤔

nihalxkumar: Kenneth's script only served as a conceptual inspiration. I have no problems mentioning this but how should I do it?

deeplow: Exactly. That's in the sense I meant. Maybe:

```suggestion
| Ben Grande <ben at invisiblethingslab dot com>
| Kenneth R. Rosen <kennethrrosen at proton.me>
```

ben-grande:
Hi. Could you please rebase your PR from the main branch to fix the conflicts?

doc/tools/qvm-template-upgrade.rst
nihalxkumar: How about this:

https://github.com/QubesOS/qubes-core-admin-linux/pull/213/changes#diff-3acefc53177842eb0ded6639f194362822f6d512bf9fca09f96fd24d09e158cbR96-R102

marmarek:
I tried to test manually what openQA complained: I got fedora-43-xfce template installed, and then ran `qvm-template-upgrade --template=fedora-43-xfce --new-name=fedora-44-upgraded` and it failed too. Log has:
```
2026-08-29 18:47:19,948 [Agent] Run entrypoint with args: Namespace(log='DEBUG', no_refresh=False, f
orce_upgrade=False, no_cleanup=False, leave_obsolete=False, download_only=False, version_upgrade='44
', show_output=False, quiet=False, no_progress=False, just_print_progress=False)
2026-08-29 18:47:19,950 [Agent] Selecting package manager.
2026-08-29 18:47:19,955 [Agent] Add `deltarpm=False` to /etc/dnf/dnf.conf
2026-08-29 18:47:20,382 [Agent] Using dnf5.
2026-08-29 18:47:21,871 [Agent] Running distribution version upgrade to 44.
2026-08-29 18:47:21,873 [Agent] run command: dnf clean all
2026-08-29 18:47:22,571 [Agent] command exit code: 0
2026-08-29 18:47:40,968 [Agent] dnf5 repo load for release 44 took 18.375s
2026-08-29 18:47:46,514 [Agent] dnf5 dependency resolution for release 44 took 5.543s
2026-08-29 18:47:46,515 [Agent] Distro-sync found nothing to do.
2026-08-29 18:47:46,603 [Agent] release upgrade did not reach 44; os-release reports '43'.
```
I don't have any more details yet. Maybe setting `releasever` variable didn't work?

marmarek:
The output (also seen in the openqa report) is:
```
Plan: upgrade fedora-43-xfce (fedora 43) -> clone fedora-44-upgraded (fedora 44)
Cloning fedora-43-xfce -> fedora-44-upgraded
fedora-44-upgraded: Cloning private volume
fedora-44-upgraded: Cloning root volume
Running version-upgrade agent in fedora-44-upgraded (-> 44)
fedora-44-upgraded:out: Preparing distribution upgrade; dependency calculation may take some time...
fedora-44-upgraded:out: Fetching 6 packages [0.00 B]                                                
fedora-44-upgraded:out: Fetching Qubes VM Repository (devel testing) [3.38 KiB]                     
fedora-44-upgraded:out: Fetching Fedora 44 - x86_64 [18.21 KiB]                                     
fedora-44-upgraded:out: Fetching Fedora 44 openh264 (From Cisco) - x86_64 [986.00 B]                
fedora-44-upgraded:out: Fetching Fedora 44 - x86_64 - Updates [15.82 KiB]                           
fedora-44-upgraded:out: Fetching Qubes OS Repository for VM (updates) [1.97 KiB]                    
fedora-44-upgraded:out: Fetching qubes testing [1.97 KiB]                                           
fedora-44-upgraded:out: repomd.xml GPG signature verification error: Signing key not found          
fedora-44-upgraded:out: repomd.xml GPG signature verification error: Signing key not found          
fedora-44-upgraded:out: Fetching qubes testing [1.97 KiB]                                           
fedora-44-upgraded:out: Fetching Qubes OS Repository for VM (updates) [1.97 KiB]                    
fedora-44-upgraded:out: Calculating package changes...                                              
fedora-44-upgraded (fedora 43 -> 44)   0.0% |                                              | [01:13]
Upgrade failed: in-VM version-upgrade agent failed for fedora-44-upgraded (exit code 24); see /var/log/qubes/update-fedora-44-upgraded.log
Leaving clone fedora-44-upgraded in place (--keep-new-on-failure).
error: in-VM version-upgrade agent failed for fedora-44-upgraded (exit code 24); see /var/log/qubes/update-fedora-44-upgraded.log
```
So, it clearly did set releasever, as seen in the fedora repository names.

marmarek:
Running `dnf distro-sync --relasever=44` manually got me this:
```
Failed to resolve the transaction:
Problem 1: installed package rpmfusion-free-release-43-1.noarch requires system-release(43), but none of the providers can be installed
  - fedora-release-43-27.noarch does not belong to a distupgrade repository
  - problem with installed package
 Problem 2: installed package fedora-release-43-27.noarch requires fedora-release-common = 43-27, but none of the providers can be installed
  - installed package rpmfusion-nonfree-release-43-1.noarch requires system-release(43), but none of the providers can be installed
  - fedora-release-common-43-27.noarch does not belong to a distupgrade repository
  - problem with installed package
You can try to add to command line:
  --skip-broken to skip uninstallable packages
```

So, there are two issues:
1. The qvm-template-upgrade tool (or specifically its libdnf5 handling) does not report transaction resolving failure correctly.
2. The upgrade fails due to rpmfusion release packages being installed, but related repositories being disabled.

Solving the second one might be tricky. Having rpmfusion-*-release packages installed is intentional, so users can easily enable it when needed, but it should still be up to the user to do that - that's why those repositories are disabled. The dependency there is strict, so just excluding rpmfusion packages from the upgrade won't help. What does work is temporarily enable the rpmfusion repositories (`rpmfusion-free,rpmfusion-nonfree`). The issue with this approach is that it risks fetching other packages from there too (on top of just `rpmfusion-*-release` packages), against user will. But maybe there is an option to avoid that, for example by using `includepkgs` option for those repositories? Or maybe by setting higher `priority` value for those repositories temporarily?
In any case, it would be good to also inspect transaction if no extra packages got pulled in from rpmfusion.

So, what I'd propose is:
1. Check if `rpmfusion-free-release` package is installed.
2. Check if `rpmfusion-free` repository is enabled.
3. If the first is "yes" and the second is "no", then temporarily set `rpmfusion-free` repository to `enabled=1` and `includepkgs=rpmfusion-free-release` (on the repository object, just after loading the configuration?)
4. After resolving the goal, inspect transaction if really no unexpected packages got pulled in from rpmfusion repository (ofc only if it got enabled in step 3).

Do the same for `rpmfusion-nonfree` too.

nihalxkumar:
@marmarek I tried on minimal fedora-43 template
working perfectly https://asciinema.org/a/YJ7xl4xytytAZJ7f

marmarek:
Well, yes, it worked on minimal because it doesn't have rpmfusion included...

nihalxkumar:
right. I currently have a running upgrade for fedora-43-xfce 
will share the final result and logs shortly

nihalxkumar:
successful https://asciinema.org/a/9xZ74qLWNNnRDdc2
however, the existing approach uninstalls rpmfusion-free-release-43 and rpmfusion-nonfree-release-43, instead of upgrading them

marmarek:
Ok, that's another approach, yes. Not perfect, but probably good enough for now.

In any case, it would be good to fix reporting error on transaction resolve step. I found another way you can simulate the failure - edit `/etc/yum.repos.d/qubes-r4.repo` and replace `$releasever` with hardcoded old version (like 43). This should result in an upgrade failure and you can see that no details are shown to the user.

marmarek:
Ok, now the transaction failure info is included in the log, which is good. But on the tool output it still have last status as `Calculating package changes...` and then the generic error message. Maybe add also some message there, like `Found N problems when resolving packages dependencies`, or something like this? It may be okay to leave full details only in log (they can be quite long in some cases).

Other than that, it seems to work now, and indeed removes rpmfusion package. openQA test says it works now :)
But, unit test and mypy fail in CI...

marmarek:
I see a conflict now, and that prevents CI from running :(

nihalxkumar:
Yes, I am fixing

marmarek:
@ben-grande are you happy with the current shape of this PR?

vmupdate/agent/source/apt/apt_cli.py
```diff
@@ -176,3 +194,209 @@ def remove_obsolete_kernels(self) -> ProcessResult:
...
+
+        # Refuse early if no source file references the old or new codename.
+        try:
+            sources = [
+                open(path, "r", encoding="utf-8").read()
+                for path in self._apt_source_paths()
+            ]
+        except OSError as exc:
+            return self._refuse(f"cannot read apt sources: {exc}.")
+        if not any(
```
ben-grande, line 229: I think the `any` here is fine. It shouldn't be all because some sources might not even have a proper suite, such as

- Signal: https://updates.signal.org/static/desktop/apt/signal-desktop.sources
- Element: https://element.io/en/download#linux

ben-grande: This gives me an idea, a test that includes external repos, can even be a fake/localhost one, that doesn't have a codename in the sources file, just to guarantee that this `any` is not changed to `all` in the future.

vmupdate/agent/source/apt/apt_cli.py
```diff
@@ -176,3 +194,209 @@ def remove_obsolete_kernels(self) -> ProcessResult:
...
+                f"or {new_codename!r}; refusing upgrade."
+            )
+
+        self._report_progress(0.0)
+        result = ProcessResult()
+
+        # Update before switching sources and running dist-upgrade: a stale
+        # system risks unresolvable transactions across the release boundary.
+        # Each step gets its own slice of the bar, weighted by its share of
+        # a measured debian-12 to 13 template upgrade; without that, every
```
ben-grande, line 244: Any tips on how to choose the step number in case something else needs to be added here? Please include a comment about it. As in, how it was measured to set this step configuration.

vmupdate/agent/source/apt/apt_cli.py
```diff
@@ -176,3 +194,209 @@ def remove_obsolete_kernels(self) -> ProcessResult:
...
+    @classmethod
+    def _debian_codename(cls, target: str) -> Optional[str]:
+        """Return Debian's codename for a numeric release."""
+        with open(
+            cls.DEBIAN_RELEASES_FILE,
+            "r",
+            encoding="utf-8",
+            newline="",
+        ) as releases_file:
+            for release in csv.DictReader(releases_file):
```
ben-grande, line 339: Please use `csv.reader`: https://docs.python.org/3/library/csv.html#csv.reader.

vmupdate/agent/source/common/package_manager.py
```diff
@@ -83,6 +87,46 @@ def upgrade(
...
+                postinstall = subprocess.call(
+                    ["/etc/qubes-rpc/qubes.PostInstall"]
+                )
+            except OSError as exc:
+                postinstall = -1
+                self.log.warning("could not run qubes.PostInstall: %s", exc)
+            if postinstall != 0:
+                self.log.warning(
+                    "qubes.PostInstall exited with %d; dom0 metadata "
+                    "(qvm-features) will refresh on next qube start.",
```
ben-grande, line 114: please change "dom0 metadata (qvm-features)" to "qube features".

vmupdate/agent/source/common/package_manager.py
```diff
@@ -321,6 +365,73 @@ def upgrade_internal(self, remove_obsolete: bool) -> ProcessResult:
...
+        raise NotImplementedError(
+            "Distribution version upgrade is not implemented for this "
+            f"package manager ({self.package_manager})."
+        )
+
+    def _verify_release_upgrade(
+        self, target: str, os_data: dict
+    ) -> ProcessResult:
+        """Verify the target is a valid single-step upgrade from the current release."""
+        if not (target.isascii() and target.isdigit()):
```
ben-grande, line 379: Is  there a digit that is not ASCII?

vmupdate/agent/source/common/package_manager.py
```diff
@@ -321,6 +365,73 @@ def upgrade_internal(self, remove_obsolete: bool) -> ProcessResult:
...
+            return self._refuse(f"invalid target release {target!r}.")
+
+        current_major = os_data.get("release", "").split(".")[0]
+        if not (current_major.isascii() and current_major.isdigit()):
+            return self._refuse(
+                f"cannot read a numeric in-qube release from "
+                f"{os_data.get('release')!r}."
```
ben-grande, line 386: same release to a variable to avoid getting it again here and on the next block.

vmupdate/agent/source/common/package_manager.py
```diff
@@ -321,6 +365,73 @@ def upgrade_internal(self, remove_obsolete: bool) -> ProcessResult:
...
+            )
+        if int(target) != int(current_major) + 1:
+            return self._refuse(
+                f"in-qube release {os_data.get('release')!r} can only move "
```
ben-grande, line 390: https://github.com/QubesOS/qubes-core-admin-linux/pull/213/changes#diff-12d64ae6128a5eba726f07e79fbe636645196b055db7ef9b3216cd4f60751416R386

vmupdate/agent/source/common/progress_reporter.py
```diff
@@ -84,6 +84,91 @@ def _format_bytes(size: int | float) -> str:
...
+
+    Rescales package progress to 0..CAP, then asymptotically advances
+    from CAP to TAIL_STOP as post-transaction scriptlets execute.
+    TAIL_STOP < 100 because qubes.PostInstall still runs afterwards.
+
+    Mixin for Progress subclasses. Inert until open_tail() is called.
+    """
+
+    # Overall percent the package callbacks are rescaled to end at.
+    CAP = 90.0
```
ben-grande, line 98: Please add explanation of why this value was chosen, same thing for the other global variables.

vmupdate/agent/source/dnf/dnf5_api.py
```diff
@@ -257,8 +425,10 @@ def mirror_failure(
         """
         if isinstance(msg, bytes):
             msg = msg.decode("ascii", errors="ignore")
+        # libdnf5 passes no metadata type for plain package downloads
+        # (avoid "Fetching None failure (...)").
         print(
-            f"Fetching {metadata} failure "
+            f"Fetching {metadata or 'package'} failure "
```
ben-grande, line 431: I will assume that using `'package'` here is enough information because the `msg` below should contain the package name.

vmupdate/agent/source/dnf/dnf_api.py
```diff
@@ -166,6 +171,79 @@ def upgrade_internal(self, remove_obsolete: bool) -> ProcessResult:
...
+                    self.log.info("Distro-sync found nothing to do.")
+                    return result
+                download_started = time.monotonic()
+                base.download_packages(
+                    trans.install_set, progress=self.progress.fetch_progress
+                )
+                self.log.debug(
+                    "dnf package download for release %s took %.3fs",
+                    target,
+                    time.monotonic() - download_started,
```
ben-grande, line 221: I counted 13 occurrences of the pattern:

- record time
- run action
- log time difference

Can be made with [contextlib.contextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager).

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,620 @@
...
+        assert self.source_vm is not None  # set by validate()
+        self.log.info("Cloning %s -> %s", self.source_vm.name, self.new_name)
+        self.cloned_qube = self.app.clone_vm(self.source_vm, self.new_name)
+
+    @contextlib.contextmanager
+    def _disk_space_warning_suppressed(self) -> Iterator[None]:
+        """Temporarily suppress qui-disk-space during upgrade.
+
+        The upgrade briefly holds both releases' packages, which can
+        trip the 90% threshold. This may bring unnecesarry panic for users.
```
ben-grande, line 408: Hm.. 90% is above what I expected. When we discussed about having enough space in the root file system for upgrades, I thought the threshold was bigger. You reached 90% while having how much total space in the root partition?

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,620 @@
...
+    override: Optional[str],
+) -> str:
+    """Replace the version in the source name with the target version.
+
+    Examples:
+        fedora-41, 41 -> 42  =>  fedora-42
+
+        fedora-41-minimal, 41 -> 42  =>  fedora-42-minimal
+
+        custom, 41 -> 42  =>  custom-42
```
ben-grande, line 201: Not for this PR, but in the future, having an option to keep the same name when it doesn't have the version in the name, would be interesting for standalones and custom named templates.

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,620 @@
...
+# Argument parsing / logging
+
+
+def get_parser() -> qubesadmin.tools.QubesArgumentParser:
+    parser = qubesadmin.tools.QubesArgumentParser(
+        prog="qvm-template-upgrade",
+        description="Upgrade a TemplateVM or StandaloneVM to the next distro "
+        "version.",
+        # Avoid qubesadmin package metadata lookup when run from PYTHONPATH.
+        version="",
```
ben-grande, line 229: Then show the version of the vmupdate?

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,620 @@
...
+            return distro_lower, version
+        family = next(
+            (c for c in distro_like.lower().split() if c in SUPPORTED_DISTROS),
+            None,
+        )
+        if family is not None:
+            raise ValidationError(
+                f"{self.source_vm.name} is a {distro} derivative; its own "
+                f"version numbering does not match {family}'s releases, so "
+                f"it cannot be upgraded with this tool."
```
ben-grande, line 379: Examples:

- Ubuntu: has its own repos
- Kicksecure: has debian repos
- Whonix: has debian repos

I think Kicksecure and Whonix will break on this validation.

ben-grande: I still think it is important to do a test run on Kicksecure and Whonix and let us know what happens. Whonix, different than other derivatives, doesn't mess with `/etc/os-release` for example.

Imagine a whonix-gateway-18, it has the following:

```
% cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.4
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

If upgrading from whonix-gateway-12, which was bookworm, will the template be renamed to whonix-gateway-13?

ben-grande: As upgrading Whonix and Kicksecure was out of the scope, if it fails, could you please open an issue for it to be tracked?

vmupdate/template_upgrade.py
```diff
@@ -0,0 +1,620 @@
...
+        """Remove the half-upgraded clone, if any. Safe to call repeatedly."""
+        if self.cloned_qube is None:
+            return
+        self.log.warning("Removing failed clone %s", self.cloned_qube.name)
+        try:
+            # Discard the failed clone immediately; kill() no-ops when the
+            # clone has already halted.
+            try:
+                self.cloned_qube.kill()
+            except qubesadmin.exc.QubesVMNotStartedError:
```
ben-grande, line 525: When does this happen? Please document/comment.

vmupdate/template_upgrade.py
ben-grande: No need to do it, just sharing the Whonix release upgrade instructions: https://www.whonix.org/wiki/Release_Upgrade_17_to_18, not as simple as just upgrading base distributions.

ben-grande:
Hi @nihalxkumar. Are you continuing with the PR?

PR #215, opened 2026-06-09
Add CI coverage for vmupdate formatting and pylint checks.

nihalxkumar:
This extends the existing GitLab test job so vmupdate changes are checked with black and pylint after the unit tests run.

The patch also formats the current vmupdate files so the new black check starts from a clean baseline, and adds a focused pylint configuration for the checks that are actionable in this tree.

Mypy is intentionally not enabled in this PR because current upstream still has existing typing errors that need a separate cleanup before it can be a blocking CI check.

fixes: https://github.com/QubesOS/qubes-core-admin-linux/pull/213#issuecomment-4630331308

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,8 +1,16 @@
 include:
+- file: /common.yml
+  project: QubesOS/qubes-continuous-integration
 - file: /r4.3/gitlab-base.yml
   project: QubesOS/qubes-continuous-integration
 - file: /r4.3/gitlab-host.yml
   project: QubesOS/qubes-continuous-integration
+
+lint:
+  extends: .lint
+  stage: checks
+  variables:
+    DIR: vmupdate
 
 checks:tests:
   stage: checks
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,28 @@
+[MASTER]
+init-hook='import sys; sys.path.append("vmupdate/agent")'
+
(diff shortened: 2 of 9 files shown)


.gitlab-ci.yml
```diff
@@ -16,6 +16,8 @@ checks:tests:
   script:
     - export PATH="$HOME/.local/bin:$PATH"
     - PYTHONPATH=~/core-admin-client ./run-tests.sh
+    - python3 -m black -l80 --check setup.py vmupdate
```
marmarek: We have some template jobs for this already. Try:
```
include:
- file: /common.yml
  project: QubesOS/qubes-continuous-integration

...
lint:
  extends: .lint
  stage: checks
  variables:
    DIR: vmupdate
```

if pylint needs to be custom anyway, you can add `SKIP_PYLINT: 1` here and keep pylint as you do now.

nihalxkumar: Nice, this is better. 
Did as you suggested

marmarek (approved):
for now lets include it as is; but eventually I'd like to reduce number of disabled pylint checks

nihalxkumar:
feel free to ping me anytime on matrix. :)

ben-grande:
When not intending to close an issue or pull request. Use "For:" instead of "Fixes:".

nihalxkumar:
Noted, thanks

PR #218, opened 2026-07-05
Remove disabling starting updatevm under some conditions

### LINKED ISSUE #10006: qubes-update-gui fails when UpdateVM is not running and CHECK_ONLY is set
### Qubes OS release

* 4.2.4

### Brief summary

**Setup:**

In our Qubes OS configuration, we modify the default `updatevm` (referred to as `UPDATEVM`) to a VM other than `sys-firewall`. This setup is specifically designed to facilitate updates through a VPN.

To conserve resources, this dedicated `UPDATEVM` is typically not needed for general use. Therefore, we automatically shut down this VM when it's idle using `qubes-app-shutdown-idle`.

**Problem:**

We've observed that the `qubes-update-gui` application consistently fails to perform an update check for dom0 under our current setup.

The root of the problem lies in the following execution flow:

1.  The `qubes-update-gui` initiates a check for dom0 updates by running:
    ```python
    returncode = qubes_dom0_update("--refresh", "--check-only")
    ```
    This call occurs at [https://github.com/QubesOS/qubes-desktop-linux-manager/blob/3a85bee077497b234ba55490d0a7710b10a02e0c/qui/updater/progress_page.py#L183C4-L183C76](https://github.com/QubesOS/qubes-desktop-linux-manager/blob/3a85bee077497b234ba55490d0a7710b10a02e0c/qui/updater/progress_page.py#L183C4-L183C76).

2.  Subsequently, `qubes-dom0-update` encounters an `ERROR: UpdateVM not running ...` message, which leads to the termination of the dom0 update process. This behavior is dictated by the following code segment:
    ```bash
    # Do not start VM automatically when running from cron (only checking for updates)
    if [ "$CHECK_ONLY" == "1" ] && ! qvm-check -q --running "$UPDATEVM" > /dev/null 2>&1; then
        echo "ERROR: UpdateVM not running, not starting it in non-interactive mode" >&2
        exit 1
    fi
    ```
    This check is located at [https://github.com/QubesOS/qubes-core-admin-linux/blob/65eb613cd3cd3907cc6ec515f9cc16aa7d7a4851/dom0-updates/qubes-dom0-update#L249C59-L249C67](https://github.com/QubesOS/qubes-core-admin-linux/blob/65eb613cd3cd3907cc6ec515f9cc16aa7d7a4851/dom0-updates/qubes-dom0-update#L249C59-L249C67).

Since our `UPDATEVM` is often shut down for resource management, this check prevents `qubes-update-gui` from successfully performing its dom0 update check.

**Proposed Solutions / Request for Input:**

We are happy to contribute a Pull Request to address this issue but are unsure of the most appropriate solution. We see two main approaches:

1.  **Remove the `CHECK_ONLY` related VM running check in `qubes-dom0-update`:** This would allow `qubes-dom0-update` to proceed with a check even if the `UPDATEVM` is not running when `--check-only` is used.
2.  **Modify `qubes-update-gui` to always start the `UPDATEVM` before performing a dom0 update check:** This would ensure the `UPDATEVM` is active when `qubes-dom0-update` is called with `--check-only`.

Personally, we lean towards **option 1)** as it seems more aligned with the `--check-only` flag's intent. Which really should always check if there are updates and not abort under an external condition (such as updatevm not running). If the cronjob should not run when the updatevm is not running, the cronjob should handle that.

### Steps to reproduce

* shutdown UPDATEVM
* launch update using `qubes-update-gui`

### Expected behavior

*  `qubes-update-gui` should start the UPDATEVM and proceed with the update

### Actual behavior

*  `qubes-update-gui` aborts with error

### Additional information

We would greatly appreciate help and input from the Qubes OS development team on the best way to resolve this issue. Please let us know which approach is preferred or if there's an alternative solution we haven't considered.

marmarta:
When qubes-dom0-update was run with --check-only, it never started updatevm if it wasn't running. This causes problems with gui updater, but also doesn't seem to be very logical. Removed the check.

fixes QubesOS/qubes-issues#10006

### DIFF
--- a/dom0-updates/qubes-dom0-update
+++ b/dom0-updates/qubes-dom0-update
@@ -245,12 +245,6 @@ if [ "$GUI" == "1" ]; then
     fi
 fi
 
-# Do not start VM automatically when running from cron (only checking for updates)
-if [ "$CHECK_ONLY" == "1" ] && ! qvm-check -q --running "$UPDATEVM" > /dev/null 2>&1; then
-    echo "ERROR: UpdateVM not running, not starting it in non-interactive mode" >&2
-    exit 1
-fi
-
 if [ -n "$CLEAN" ]; then
     rm -f /var/lib/qubes/updates/rpm/*
     rm -f /var/lib/qubes/updates/repodata/*


marmarta:
I'm open to other ways of fixing it, btw. It just looks like this check was True for more than just "running in cron".

marmarek:
This will have consequences on daily updates check. When user stops updatevm, with this change the daily updates check will start it again, which may even happen just after user stopped it. This may be at least surprising...

PR #219, opened 2026-07-24
Lint with mypy on CI

ben-grande:
I'd like mypy to run on https://github.com/QubesOS/qubes-core-admin-linux/pull/213.

### DIFF
--- a/vmupdate/agent/source/dnf/dnf5_api.py
+++ b/vmupdate/agent/source/dnf/dnf5_api.py
@@ -164,10 +164,10 @@ class FetchProgress(DownloadCallbacks, Progress):
     def __init__(self, weight: int, log):
         DownloadCallbacks.__init__(self)
         Progress.__init__(self, weight, log)
-        self.bytes_to_fetch = 0
-        self.bytes_fetched = 0
-        self.package_bytes = {}
-        self.package_names = {}
+        self.bytes_to_fetch = 0.0
+        self.bytes_fetched = 0.0
+        self.package_bytes: dict[int, float] = {}
+        self.package_names: dict[int, str] = {}
         self.count = 0
         self.fetching_notified = False
 
@@ -267,9 +267,9 @@ class UpgradeProgress(TransactionCallbacks, Progress):
     def __init__(self, weight: int, log):
         TransactionCallbacks.__init__(self)
         Progress.__init__(self, weight, log)
-        self.pgks = None
-        self.pgks_done = None
-        self.processed_packages = set()
+        self.pgks: int | None = None
+        self.pgks_done: int | None = None
+        self.processed_packages: set[str] = set()
 
     def install_progress(
         self, item: libdnf5.base.TransactionPackage, amount: int, total: int
@@ -287,6 +287,8 @@ class UpgradeProgress(TransactionCallbacks, Progress):
             print(f"Installing {package}", flush=True)
             self.processed_packages.add(package)
         pkg_progress = amount / total
+        assert isinstance(self.pgks_done, int)
+        assert isinstance(self.pgks, int)
         percent = (self.pgks_done + pkg_progress) / self.pgks * 100
         self.notify_callback(percent)
 
@@ -314,6 +316,8 @@ class UpgradeProgress(TransactionCallbacks, Progress):
             print(f"Uninstalling {package}", flush=True)
             self.processed_packages.add(package)
         pkg_progress = amount / total
+        assert isinstance(self.pgks_done, int)
+        assert isinstance(self.pgks, int)
         percent = (self.pgks_done + pkg_progress) / self.pgks * 100
         self.notify_callback(percent)
 
--- a/vmupdate/agent/source/dnf/dnf5_api.py
+++ b/vmupdate/agent/source/dnf/dnf5_api.py
@@ -21,10 +21,12 @@
 # USA.
 
 import subprocess
+from logging import Logger, Handler
+from typing import Any
 
 import libdnf5
 from libdnf5.repo import DownloadCallbacks
-from libdnf5.rpm import TransactionCallbacks
+from libdnf5.rpm import TransactionCallbacks, Nevra
 from libdnf5.base import Goal
 
 from source.common.process_result import ProcessResult
@@ -42,7 +44,9 @@ class TransactionError(RuntimeError):
 class DNF5(DNFCLI):
     PROGRESS_REPORTING = True
 
-    def __init__(self, log_handler, log_level, agent_type: AgentType):
+    def __init__(
+        self, log_handler: Handler, log_level: int, agent_type: AgentType
+    ) -> None:
         super().__init__(log_handler, log_level, agent_type)
 
         self.base = libdnf5.base.Base()
@@ -161,7 +165,7 @@ class DNF5(DNFCLI):
 
 
 class FetchProgress(DownloadCallbacks, Progress):
-    def __init__(self, weight: int, log):
+    def __init__(self, weight: int, log: Logger) -> None:
         DownloadCallbacks.__init__(self)
         Progress.__init__(self, weight, log)
         self.bytes_to_fetch = 0.0
@@ -172,7 +176,7 @@ class FetchProgress(DownloadCallbacks, Progress):
         self.fetching_notified = False
 
     def add_new_download(
-        self, _user_data, description: str, total_to_download: float
+        self, _user_data: Any, description: str, total_to_download: float
     ) -> int:
         """
         Notify the client that a new download has been created.
@@ -264,7 +268,7 @@ class FetchProgress(DownloadCallbacks, Progress):
 
 
 class UpgradeProgress(TransactionCallbacks, Progress):
-    def __init__(self, weight: int, log):
+    def __init__(self, weight: int, log: Logger) -> None:
         TransactionCallbacks.__init__(self)
         Progress.__init__(self, weight, log)
         self.pgks: int | None = None
@@ -273,7 +277,7 @@ class UpgradeProgress(TransactionCallbacks, Progress):
 
     def install_progress(
         self, item: libdnf5.base.TransactionPackage, amount: int, total: int
-    ):
+    ) -> None:
         r"""
         Report the package installation progress periodically.
 
@@ -292,7 +296,7 @@ class UpgradeProgress(TransactionCallbacks, Progress):
         percent = (self.pgks_done + pkg_progress) / self.pgks * 100
         self.notify_callback(percent)
 
-    def transaction_start(self, total: int):
+    def transaction_start(self, total: int) -> None:
         r"""
         Preparation phase has started.
 
@@ -303,7 +307,7 @@ class UpgradeProgress(TransactionCallbacks, Progress):
 
     def uninstall_progress(
         self, item: libdnf5.base.TransactionPackage, amount: int, total: int
-    ):
+    ) -> None:
         """
         Report the package removal progress periodically.
 
@@ -322,7 +326,9 @@ class UpgradeProgress(TransactionCallbacks, Progress):
         self.notify_callback(percent)
 
     # pylint: disable=unused-argument
-    def elem_progress(self, item, amount: int, total: int):
+    def elem_progress(
+        self, item: libdnf5.base.TransactionPackage, amount: int, total: int
+    ) -> None:
         r"""
         The installation/removal process for the item has started
 
@@ -338,8 +344,11 @@ class UpgradeProgress(TransactionCallbacks, Progress):
 
     # pylint: disable=unused-argument,redefined-builtin
     def script_start(
-        self, item: libdnf5.base.TransactionPackage, nevra, type: int
-    ):
+        self,
+        item: libdnf5.base.TransactionPackage,
+        nevra: libdnf5.rpm.Nevra,
+        type: int,
+    ) -> None:
         r"""
         Execution of the rpm scriptlet has started
 
--- a/vmupdate/update_manager.py
+++ b/vmupdate/update_manager.py
@@ -45,7 +45,7 @@ class UpdateManager:
     Update multiple qubes simultaneously.
     """
 
-    def __init__(self, qubes, args, log, dom0=False):
+    def __init__(self, qubes, args, log, dom0=False) -> None:
         self.qubes = qubes
         self.max_concurrency = args.max_concurrency
         self.show_output = args.show_output
@@ -60,7 +60,7 @@ class UpdateManager:
         self.log = log
         self.dom0 = dom0
 
-    def run(self, agent_args):
+    def run(self, agent_args) -> tuple[int, dict]:
         """
         Run simultaneously `update_qube` for all qubes as separate processes.
         """
@@ -104,7 +104,7 @@ class UpdateManager:
             if qube.klass == "AdminVM" and show_progress:
                 # progress of AdminVM is continuation of different process,
                 # so we want to skip 0 value at beginning
-                progress_bar.progress_bars[qube.name].progress = None
+                progress_bar.progress_bars[qube.name].reset()
 
         progress_bar.pool.close()
         progress_bar.feeding()
@@ -130,7 +130,7 @@ class UpdateManager:
 
         return self.ret_code, progress_bar.statuses
 
-    def collect_result(self, result_tuple: Tuple[str, ProcessResult]):
+    def collect_result(self, result_tuple: Tuple[str, ProcessResult]) -> None:
         """
         Callback method to process `update_qube` output.
         """
@@ -152,7 +152,7 @@ class UpdateManager:
         elif not self.quiet and self.no_progress:
             self.print(result.out)
 
-    def print(self, *args):
+    def print(self, *args) -> None:
         if self.buffered:
             self.buffer += " ".join(map(str, args)) + "\n"
         else:
@@ -164,10 +164,10 @@ class TerminalMultiBar:
     Handles multiple progress bars in terminal.
     """
 
-    def __init__(self):
-        self.progresses = []
+    def __init__(self) -> None:
+        self.progresses: list[SimpleTerminalBar] = []
 
-    def print(self):
+    def print(self) -> None:
         for progress in self.progresses:
             print(progress, file=sys.stderr, flush=True)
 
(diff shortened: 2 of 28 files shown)


ben-grande:
@nihalxkumar You modified some of these files in your PR, I think that it is easier:

- We enable CI in your PR
- You check what new issues were introduced and fix them in the same PR

Rather than fixing those errors in a separate PR, as that might cause some merge conflicts.

nihalxkumar:
I ran the mypy check on my branch. It reports the same 34 errors already present on main, with no new errors from my changes
I think we should use #219 to fix the baseline errors before this CI job is brought into qvm-template-upgrade branch

ben-grande:
> with no new errors from my changes

Bodies of untyped functions are not checked, so you can add return type to functions you added to make mypy lint them.

nihalxkumar:
🙋 [`01027e3` (This PR)](https://github.com/QubesOS/qubes-core-admin-linux/pull/219/changes/01027e3b607d6233a5fa6e8f8a6ab032963e68c6#diff-9066f441738a455fbec81434dc6afd4bd9bf3e5add40d2698f66b8e90bcd0ccfR431) L431 renames the dom0 list to `entrypoint_cmd`, but L463 still passes `entrypoint`.
on the AdminVM + dom0 path it's unassigned and `qubes-vm-update --dom0` raises `UnboundLocalError`.

My patch for 213: https://gist.github.com/nihalxkumar/b5d76eb71ba8ebaa45ac90af08d6e9b2 
will commit and rebase after this PR lands

ben-grande:
> 🙋 [`01027e3` (This PR)](https://github.com/QubesOS/qubes-core-admin-linux/pull/219/changes/01027e3b607d6233a5fa6e8f8a6ab032963e68c6#diff-9066f441738a455fbec81434dc6afd4bd9bf3e5add40d2698f66b8e90bcd0ccfR431) L431 renames the dom0 list to `entrypoint_cmd`, but L463 still passes `entrypoint`.
> on the AdminVM + dom0 path it's unassigned and `qubes-vm-update --dom0` raises `UnboundLocalError`.

I was trying to circumvent a mypy error that entrypoint had two types, `str` and `list`. I now see that I missed one step.

vmupdate/agent/source/dnf/dnf5_api.py
```diff
@@ -166,13 +166,13 @@ def __init__(self, weight: int, log):
...
-        self.package_bytes = {}
-        self.package_names = {}
+        self.package_bytes: dict[int, int] = {}
+        self.package_names: dict[int, str] = {}
         self.count = 0
         self.fetching_notified = False
 
     def add_new_download(
-        self, _user_data, description: str, total_to_download: float
+        self, _user_data, description: str, total_to_download: int
```
marmarek: This (and other changes to function signatures in this class) is wrong - the function is called by libdnf and according to its documentation it uses float here:
```
 |  add_new_download(self, user_data, description, total_to_download)
 |      Notify the client that a new download has been created.
 |      :type user_data: void
 |      :param user_data: User data entered together with url/package to download.
 |      :type description: string
 |      :param description: The message describing new download (url/packagename).
 |      :type total_to_download: float
 |      :param total_to_download: Total number of bytes to download.
 |      :rtype: void
 |      :return: Associated user data for new download.
```

marmarek (changes requested):

vmupdate/agent/source/dnf/dnf5_api.py
ben-grande: Thanks. Was this automatic detection somehow that I could have done via editor or some other tool? To fix, I looked at `help(libdnf5.repo.DownloadCallbacks.add_new_download)`.

marmarek: Nope, I just checked type changes in this PR that looked suspicious to me.

vmupdate/update_manager.py
```diff
@@ -420,17 +423,18 @@ def _run_agent(
...
-                entrypoint = ["sudo", "qubes-dom0-update", "-y"]
+                entrypoint_cmd = ["sudo", "qubes-dom0-update", "-y"]
                 if agent_args.just_print_progress or self.show_progress:
-                    entrypoint.append("--just-print-progress")
+                    entrypoint_cmd.append("--just-print-progress")
                 if agent_args.quiet:
                     # silent is equivalent to quiet for dom0-update
-                    entrypoint.append("--silent")
+                    entrypoint_cmd.append("--silent")
+                entrypoint = " ".join(entrypoint_cmd)
```
marmarek: Here, you changed what is actually set to `entrypoint`. It used to be a list, now you set str joined from that list. Look at `QubeConnection.run_entrypoint()` - it behaves differently depending on type, and I'm pretty sure it won't work this way (it will try to run `python3 sudo qubesd-dom0-update ...`).

marmarek (changes requested):

ben-grande:
As apparently I was doing bad a type hint, I took this opportunity to train and do it on vmupdate completely (except tests). I did not enjoy it at all, but I think I did better this time.

vmupdate/update_manager.py
marmarek: This is still an issue in the latest version (at fe9f5a2)

marmarek:
And also CI is fully red (pylint, tests, mypy)

ben-grande:
Had issues annotating `status_notifier`, there doesn't seem to be a type for AutoProxy.

vmupdate/update_manager.py
```diff
@@ -180,14 +196,17 @@ class SimpleTerminalBar:
...
+    def __init__(
+        self, total: Optional[int | float], position: int, desc: Optional[str]
+    ) -> None:
+        assert SimpleTerminalBar.PARENT_MULTI_BAR is not None
         assert position == len(SimpleTerminalBar.PARENT_MULTI_BAR.progresses)
         SimpleTerminalBar.PARENT_MULTI_BAR.progresses.append(self)
-        self.desc = desc
-        self.progress = 0
+        self.desc: str = desc
+        self.n: float | None = 0.0
```
marmarek: Why `n` (it's really not more descriptive than `progress`), and also why pylint doesn't complain about too short attribute name?

vmupdate/update_manager.py
```diff
@@ -104,7 +116,11 @@ def run(self, agent_args):
             if qube.klass == "AdminVM" and show_progress:
                 # progress of AdminVM is continuation of different process,
                 # so we want to skip 0 value at beginning
-                progress_bar.progress_bars[qube.name].progress = None
+                qube_bar = progress_bar.progress_bars[qube.name]
+                progress = None
+                if isinstance(qube_bar, tqdm):
+                    progress = 0.0
```
marmarek: Why, or more specifically - why here? Changing behavior based on what progress class is used (outside of that class) looks like a footgun, in case somebody wants to implement different progress frontend.

marmarek (changes requested):
Besides specific comments, is it a good idea to add this rather big PR, before #213 lands? Or is @nihalxkumar okay with potentially painful rebase?

vmupdate/update_manager.py
marmarek: Oh, I see why, pylintrc has `disable=invalid-name` ...

ben-grande: > Why `n` (it's really not more descriptive than `progress`)

In the commit message, I wrote this:

_The progress bar variable changed from "progress" to "n" as that is what "tqdm" has._

The type hint was complaining that `tqdm` doesn't have progress attribute, and on my tests, it can't be set to `None`. I preferred using the same attribute name, although a very short single letter. If you prefer the long name, I will have to make sure that all writes to `.progress` attribute are made to `.n` if terminal bar is `tqdm`.

vmupdate/update_manager.py
ben-grande: answered on the comment above why. About why here, no specific reason, it's just the only place where it was needed. I will see about making a method to avoid accessing the attribute directly, and setting to a value is that not valid depending on the terminal bar.

marmarek: The code comment just above suggests the `None` value was important. Is it not for tqdm?

ben-grande: `tqdm` doesn't accept `None` as a value:

```python3
>>> import tqdm
>>> help(tqdm.tqdm)
>>> t = tqdm.tqdm()
0it [00:00, ?it/s]>>>
>>> vars(t)
{'iterable': None, 'desc': '', 'total': None, 'leave': True, 'fp': <tqdm.utils.DisableOnWriteError object at 0x7c918c388050>, 'ncols': 120, 'nrows': 19, 'mininterval': 0.1, 'maxinterval': 10.0, 'miniters': 0, 'dynamic_miniters': True, 'ascii': False, 'disable': False, 'unit': 'it', 'unit_scale': False, 'unit_divisor': 1000, 'initial': 0, 'lock_args': None, 'delay': 0.0, 'gui': False, 'dynamic_ncols': False, 'smoothing': 0.3, '_ema_dn': <tqdm.std.EMA object at 0x7c918c3881a0>, '_ema_dt': <tqdm.std.EMA object at 0x7c918c47b110>, '_ema_miniters': <tqdm.std.EMA object at 0x7c918c47b250>, 'bar_format': None, 'postfix': None, 'colour': None, '_time': <built-in function time>, 'last_print_n': 0, 'n': 0, 'pos': 0, 'sp': <function tqdm.status_printer.<locals>.print_status at 0x7c918c382200>, 'last_print_t': 1785770639.7390816, 'start_t': 1785770639.7390816}
>>> t.total = 300
>>> t.n = 50
>>> t.display()
 17%|█████████████▋                                                                    | 50/300 [00:31<02:35,  1.61it/s]True
>>> 50 / 300
0.16666666666666666
>>> t.n = None
>>> t.display()
  File "<python-input-11>", line 1, in <module>
    t.display()
    ~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/tqdm/std.py", line 1495, in display
    self.sp(self.__str__() if msg is None else msg)
            ~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/tqdm/std.py", line 1151, in __str__
    return self.format_meter(**self.format_dict)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/tqdm/std.py", line 534, in format_meter
    if total and n >= (total + 0.5):  # allow float imprecision (#849)
                 ^^^^^^^^^^^^^^^^^^
TypeError: '>=' not supported between instances of 'NoneType' and 'float'
```

marmarek: But this assumes `tqdm.n` is what needs to be set. This is absolutely not clear for me from reading the current code. tqdm's documentation never mentions this attribute explicitly, probably changing it behind its backs is a wrong idea (looks like it's supposed to be done via `tqdm.update()` method, which is called in `MultipleUpdateMultipleProgressBar._update()`).

ben-grande: Thanks, this helped me notice a missing method, `reset()`.

ben-grande:
> Besides specific comments, is it a good idea to add this rather big PR, before #213 lands? Or is @nihalxkumar okay with potentially painful rebase?

I tested merging #213 on top of this PR, there was a conflict on two files, two conflicts per file. All were short and easy to solve. I tested rebase, the conflicts were spread across 3 commits.

marmarek:
> * TC_00_Dom0Upgrade_debian-13-xfce: [test_000_update_qubes_vm_update](https://openqa.qubes-os.org/tests/191674#step/TC_00_Dom0Upgrade_debian-13-xfce/2) (failure)
>       `~~~~~~~~~^^^^^^^^^^^^^^^^^^... AssertionError: qubes-vm-update failed:`

This looks to be caused by this PR. `qubes-vm-update --targets=dom0` prints not much details (just it got some exception, but not even exception name), but after adding strategical `traceback.format_exc()` in vmupdate/update_manager.py line 407, I got this:
```
dom0:out: ERROR (exception Traceback (most recent call last):
dom0:out:   File "/usr/lib/python3.13/site-packages/vmupdate/update_manager.py", line 399, in update_qube
dom0:out:     result = runner.run_agent(
dom0:out:         agent_args=agent_args,
dom0:out:         status_notifier=status_notifier,
dom0:out:         termination=termination,
dom0:out:     )
dom0:out:   File "/usr/lib/python3.13/site-packages/vmupdate/update_manager.py", line 468, in run_agent
dom0:out:     result = self._run_agent(agent_args, new_status_notifier, termination)
dom0:out:   File "/usr/lib/python3.13/site-packages/vmupdate/update_manager.py", line 510, in _run_agent
dom0:out:     assert src_dir is not None
dom0:out:            ^^^^^^^^^^^^^^^^^^^
dom0:out: AssertionError
dom0:out: )
```

vmupdate/update_manager.py
```diff
@@ -402,35 +451,43 @@ def __init__(self, app, qube, agent_args, show_progress, dom0):
...
-                entrypoint = ["sudo", "qubes-dom0-update", "-y"]
+                entrypoint_cmd = ["sudo", "qubes-dom0-update", "-y"]
                 if agent_args.just_print_progress or self.show_progress:
-                    entrypoint.append("--just-print-progress")
+                    entrypoint_cmd.append("--just-print-progress")
                 if agent_args.quiet:
                     # silent is equivalent to quiet for dom0-update
-                    entrypoint.append("--silent")
+                    entrypoint_cmd.append("--silent")
+                entrypoint: list[str] | str = entrypoint_cmd
```
marmarek, line 490: With keeping it a list, secondary variable isn't really needed, please restore to using just `entrypoint` (just adding type hint should be enough, if  mypy can't figure it out on its own).

ben-grande:
```
update_manager.py:433: error: Item "str" of "list[str] | str" has no attribute "append"  [union-attr]
update_manager.py:436: error: Item "str" of "list[str] | str" has no attribute "append"  [union-attr]
```

I added two variables to deal with this. I thought it was better to have a type per variable instead of using `# type: ignore`.

marmarek: What if you let it be just `list[str]` in this code branch? It should still be compatible with `list[str] | str` later, no?

marmarek: But if that doesn't work, then fine, lets have a separate variable.

ben-grande:
> dom0:out: ERROR (exception Traceback (most recent call last):
> dom0:out:   File "/usr/lib/python3.13/site-packages/vmupdate/update_manager.py", line 399, in update_qube
> dom0:out:     result = runner.run_agent(
> dom0:out:         agent_args=agent_args,
> dom0:out:         status_notifier=status_notifier,
> dom0:out:         termination=termination,
> dom0:out:     )
> dom0:out:   File "/usr/lib/python3.13/site-packages/vmupdate/update_manager.py", line 468, in run_agent
> dom0:out:     result = self._run_agent(agent_args, new_status_notifier, termination)
> dom0:out:   File "/usr/lib/python3.13/site-packages/vmupdate/update_manager.py", line 510, in _run_agent
> dom0:out:     assert src_dir is not None
> dom0:out:            ^^^^^^^^^^^^^^^^^^^
> dom0:out: AssertionError
> dom0:out: )

This is strange. I don't know what used to happen before, because:

- [it defaults to None](https://github.com/QubesOS/qubes-core-admin-linux/blob/4973653e85483f253d4a48ac53c37703f0372dd8/vmupdate/update_manager.py#L424)
- [only if qube klass is not AdminVM, it sets `src_dir`](https://github.com/QubesOS/qubes-core-admin-linux/blob/4973653e85483f253d4a48ac53c37703f0372dd8/vmupdate/update_manager.py#L442)
- [it can't be `None` on qube_connection](https://github.com/QubesOS/qubes-core-admin-linux/blob/4973653e85483f253d4a48ac53c37703f0372dd8/vmupdate/qube_connection.py#L127)

It can be empty string though:

```python3
>>> import os
>>> import tempfile
>>> arch_format = ".tar.gz"
>>> arch_dir = tempfile.mkdtemp()
>>> root_dir = os.path.dirname(None)
  File "<python-input-4>", line 1, in <module>
    root_dir = os.path.dirname(None)
  File "<frozen posixpath>", line 178, in dirname
TypeError: expected str, bytes or os.PathLike object, not NoneType
>>> root_dir = os.path.dirname("")
>>> root_dir
''
>>> base_dir = os.path.basename("".strip(os.sep))
>>> base_dir
''
>>> src_arch = os.path.join(arch_dir, base_dir + arch_format)
>>> src_arch
'/tmp/tmpxxt_4vja/.tar.gz'
```

But this is really weird.

vmupdate/update_manager.py
ben-grande: > What if you let it be just `list[str]` in this code branch? It should still be compatible with `list[str] | str` later, no?

Mypy doesn't like setting type twice for a variable.

```
update_manager.py:439: error: Name "entrypoint" already defined on line 431  [no-redef]
```

marmarek:
> [it can't be `None` on qube_connection](https://github.com/QubesOS/qubes-core-admin-linux/blob/4973653e85483f253d4a48ac53c37703f0372dd8/vmupdate/qube_connection.py#L127)

But that's `transfer_agent` function, which isn't called for AdminVM. So, in case of AdminVM, None is fine

ben-grande:
openQArun TEST=system_tests_network_updates TEST_TEMPLATES=debian-13-xfce UPDATE_TEMPLATES=debian-13-xfce

---

Ok, this repo is not in the allow list...

marmarek:
Regarding last push, you forgot to remove old assert location

ben-grande:
Do you know what failed here? https://github.com/QubesOS/qubes-core-admin/pull/783#issuecomment-5216580976

marmarek:
It "worked" (but see the failure there, merge conflict): https://gitlab.com/QubesOS/qubes-continuous-integration/-/pipelines/2740561055
But, github has some silly limits, setting deployment status failed:
```
{"message":"Validation Failed","errors":[{"resource":"DeploymentStatus","code":"custom","message":"This deployment has reached the maximum number of statuses."}],"documentation_url":"https://docs.github.com/rest/deployments/statuses#create-a-deployment-status","status":"422"}
```

ben-grande:
> It "worked" (but see the failure there, merge conflict):

Ok, github broke the search syntax

https://doc.qubes-os.org/en/latest/developer/debugging/automated-tests.html#add-label-to-your-pr-on-github

`org:QubesOS is:pr is:open label:openqa-group-1`

https://github.com/search?q=org%3AQubesOS+is%3Apr+label%3Aopenqa-group-2+is%3Aopen&type=code

> Some qualifiers in your query (is:pr, label) are not supported when searching code. [Try searching for pull requests instead?](https://github.com/search?q=org%3AQubesOS+is%3Apr+label%3Aopenqa-group-2+is%3Aopen&type=pullrequests&query=org%3AQubesOS+is%3Apr+label%3Aopenqa-group-2+is%3Aopen)

The new URL is:

https://github.com/search?q=org%3AQubesOS+is%3Apr+label%3Aopenqa-group-2+is%3Aopen&type=pullrequests&query=org%3AQubesOS+is%3Apr+label%3Aopenqa-group-2+is%3Aopen

And if you notice, the `type=pullrequests` is not part of the query, requiring a click on the link, rather than allowing us to provide a human readable search syntax, which I expected to have less change of breaking.

marmarek (approved):

PR #221, opened 2026-08-05
vmupdate: apply whonix settings when downloading dom0 updates

### LINKED ISSUE #11034: updating dom0 via whonix fails with qubes-vm-update
## Observation

openQA test in scenario qubesos-4.3-pull-requests-x86_64-system_tests_network_updates@64bit fails in
[TC_00_Dom0Upgrade_whonix-gateway-18](https://openqa.qubes-os.org/tests/191034/modules/TC_00_Dom0Upgrade_whonix-gateway-18/steps/2)

```
# test_000_update_qubes_vm_update
# failure: 

# timestamp 2026-08-04T03:52:35.151053
  File "/usr/lib64/python3.13/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/dom0_update.py", line 404, in test_000_update_qubes_vm_update
    self._test_000_update(self.update_func_qubes_vm_update)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/dom0_update.py", line 370, in _test_000_update
    do_update_func()
    ~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/qubes/tests/integ/dom0_update.py", line 353, in update_func_qubes_vm_update
    self.fail("qubes-vm-update failed: " + f_log.read())
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: qubes-vm-update failed: 
dom0 (pending):   0%|          | 0/100 [00:00<?, ?it/s]
dom0 (updating):   0%|          | 0/100 [00:23<?, ?it/s]
dom0 (updating):   0%|          | 0.0/100 [00:23<?, ?it/s]
dom0 (updating):   0%|          | 0.0/100 [00:23<?, ?it/s]
dom0 (error):   1%|          | 1.11/100 [00:23<35:20, 21.44s/it]
dom0 (error):   1%|          | 1.11/100 [00:23<35:21, 21.45s/it]
dom0:err: Using test-inst-updatevm as UpdateVM for Dom0
dom0:err: Downloading updates. This may take a while...
dom0:out: [0;31m[0mdom0:out: Refreshing package info
dom0:out: dom0:out: Refreshing available packages.
dom0:out: dom0:err: (process:3990): librepo-WARNING **: 03:52:04.957: LRO_METALINKURL processing failed: Curl error (6): Could not resolve hostname for https://mirrors.fedoraproject.org/metalink?repo=fedora-41&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]
dom0:out: dom0:err: Errors during downloading metadata for repository 'fedora':
dom0:out: dom0:err:   - Curl error (6): Could not resolve hostname for https://mirrors.fedoraproject.org/metalink?repo=fedora-41&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]
dom0:out: dom0:err: Failed to download metadata for repo 'fedora': Cannot prepare internal mirrorlist: Curl error (6): Could not resolve hostname for https://mirrors.fedoraproject.org/metalink?repo=fedora-41&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]
dom0:out: test-inst-updatevm:out: 
dom0:out: test-inst-updatevm:err: 
dom0:err: dom0 done error
dom0:out: 
dom0:err: 
```

## Reproducible

Fails since (at least) Build [2026073003-devel](https://openqa.qubes-os.org/tests/190411)

## Expected result

Last good: [2026072902-devel](https://openqa.qubes-os.org/tests/190341) (or more recent)

## Further details

Always latest result in this scenario: [latest](https://openqa.qubes-os.org/tests/latest?arch=x86_64&distri=qubesos&flavor=pull-requests&machine=64bit&test=system_tests_network_updates&version=4.3)

marmarek:
Whonix Gateway adds a wrapper for /usr/bin/dnf, which adds proxy-related
options. This works fine as long as it's called as a standalone process,
but when qubes-vm-update calls DNF via Python API, the wrapper is
not used, and the options are not set. This way even generic uwtwrapper
is not used, which makes DNF attept connections directly (without any
proxy), which is (intentionally) blocked.

Fix this by adding the dnf's uwt wrapper logic into vmupdate code,
when it's using API. This basically covers two parts:
 - waiting for Tor daemon to be running
 - configuring dnf to use Tor as a proxy

Fixes QubesOS/qubes-issues#11034

### DIFF
--- a/vmupdate/agent/source/dnf/dnf5_api.py
+++ b/vmupdate/agent/source/dnf/dnf5_api.py
@@ -49,6 +49,7 @@ class DNF5(DNFCLI):
         conf = self.base.get_config()
 
         if self.type == AgentType.UPDATE_VM:
+            self.configure_whonix_maybe(conf)
             conf.config_file_path = (
                 self.UPDATE_VM_INSTALLROOT + "/etc/dnf/dnf.conf"
             )
--- a/vmupdate/agent/source/dnf/dnf_api.py
+++ b/vmupdate/agent/source/dnf/dnf_api.py
@@ -50,6 +50,7 @@ class DNF(DNFCLI):
         conf.read(filename=dnfconf)
 
         if self.type == AgentType.UPDATE_VM:
+            self.configure_whonix_maybe(conf)
             conf.best = True
             conf.plugins = False
             conf.installroot = self.UPDATE_VM_INSTALLROOT
--- a/vmupdate/agent/source/dnf/dnf_cli.py
+++ b/vmupdate/agent/source/dnf/dnf_cli.py
@@ -21,6 +21,10 @@
 
 import shutil
 from typing import List
+import os
+import random
+import string
+import subprocess
 
 from source.common.package_manager import PackageManager, AgentType
 from source.common.process_result import ProcessResult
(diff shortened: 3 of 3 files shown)


marmarek:
@adrelanos @arraybolt3 take a look at this (and the linked issue). This PR basically duplicates https://github.com/Whonix/uwt/blob/master/usr/bin/dnf-3.anondist wrapper. I'm not a big fan of it (layering violation), but also I don't have much better idea. For the proxy config part, I guess Whonix could drop some config snippet in addition to / instead of --setopt in the wrapper; but that doesn't help with the "wait for tor start" part...
Do you have some better idea?

ArrayBolt3:
Unless there's some way to hook into DNF's API with something that runs arbitrary code before any action that does network access, I don't think there probably is a better way. If I remember correctly, I think there was some specific reason we used `--setopt` parameters rather than configuration files (perhaps because configuration files are too easy to unintentionally override?), but my memory on this is hazy. @adrelanos do you remember why we use options instead of config settings here?

I don't see anything wrong with the implementation. The only thing that initially seemed worrying is that a non-CSPRNG is being used to create the password, but I *really* doubt that's needed here, all that matters is that the password is unique compared to previous runs and even a simple counter could pull that off. So IMO this looks good.

marmarek:
> I think there was some specific reason we used `--setopt` parameters rather than configuration files (perhaps because configuration files are too easy to unintentionally override?), but my memory on this is hazy.

Hm, I don't remember either. But it might be something about the dom0 update script forcing alternative config (copied from dom0) instead of using the default one.

> Unless there's some way to hook into DNF's API with something that runs arbitrary code before any action that does network access,

Actually, there might be. We use DNF plugins for notifying dom0 about template update status on Fedora, something like this could be used here maybe?
The pluigin code is here: https://github.com/QubesOS/qubes-core-agent-linux/blob/main/package-managers/dnf-qubes-hooks.py and the API documentation is at https://dnf.readthedocs.io/en/latest/api_plugins.html
That's for DNF5. There is also DNF5 (not in trixie, but seems to be in forky already), which has completely different plugins API (since it's written in C++, not Python). For that, there is already a plugin that should (hopefully) be flexible enough: https://dnf.readthedocs.io/en/latest/api_plugins.html, it would be just about configuring it.

And with all that, given the config path case above, it would still need pointing at some specific plugins directory (or simply ensuring it would load plugins from default directory, instead of the alternative one).

Honestly, after looking at this, maybe the extra complexity isn't worth it and the current PR isn't that bad...

PR #225, opened 2026-09-03
waawvmupdate: run qubes.PostUpdate hook after every update attempt

### LINKED ISSUE #10389: add update hook RPC(s) in addition to qubes.PostInstall

It would be useful to be able to add custom functionality to the process of updating templates. For example, when flatpaks are installed in templates, it would be nice for `qubes-update-gui` to also update the flatpaks after the system package manager updates.

Currently, some guides and salt repos approach this by adding a script to `/etc/qubes/post-install.d` (see [Solene's flatpak guide](https://forum.qubes-os.org/t/handle-flatpak-from-salt/30157) or [my salt state definition](https://codeberg.org/devkate/spiro-fueled-kimchi-binge/src/branch/main/salt/common/flatpak/init.sls)). The problem is that the `qubes.PostInstall` RPC is only called by the VM update agent when at least one package was upgraded ([source](https://github.com/QubesOS/qubes-core-admin-linux/blob/631d1612032b7e113b596a9223f52ddb45357ad7/vmupdate/agent/source/dnf/dnf5_api.py#L121)). This is a more significant issue when using minimal templates, as the time between system package upgrades can be longer.

### The solution you'd like

In addition to the `qubes.PostInstall` hook (which [appears](https://dev.qubes-os.org/projects/core-admin/en/latest/qubes-features.html#qvm-features-request-qubes-postinstall-service) primarily designed for installed packages to request qube features), I would like there to be a `qubes.PostUpdate` RPC which is called after the system package manager upgrades, regardless of whether any system packages were upgraded. A `qubes.PreUpdate` could be useful as well, but is not necessary for my use case.

### The value to a user and who that user might be

Using these RPC hooks, users can use other package managers (flatpak, pip, npm, etc.) in templates and ensure these are kept up to date.

### Completion criteria checklist

(This section is for developer use only. Please do not modify it.)

andreasglashauser:
The dnf agents return before the transaction when there is nothing to upgrade, and apt-get never invokes dpkg (and so never `DPkg::Post-Invoke`) when there is nothing to do. As a result `qubes.PostInstall` and any `post-install.d` script only run after a real package change, and secondary package managers hooked there lag behind on quiet templates like debian minimal.

After the package manager finishes with `OK` or `OK_NO_UPDATES` in a template or standalone qube, the agent now runs `/etc/qubes-rpc/qubes.PostUpdate`, which executes` /etc/qubes/post-update.d/*.sh `regardless of whether any package changed. The script itself ships with QubesOS/qubes-core-agent-linux#676. If it is missing, the template runs an older core-agent and the step is skipped.

Fixes QubesOS/qubes-issues#10389.

AI disclosure: I used a local LLM for writing commit messages. I have written each line of code myself, but I used Claude Fable 5 to brainstorm an implementation plan and review my code. Fable also edited `doc/tools/qubes-vm-update.rst` for me.


vmupdate/agent/source/common/post_update_hook.py
```diff
@@ -0,0 +1,82 @@
...
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    ) as proc:
+        result = ProcessResult.process_communicate(proc)
+    if result.code != EXIT.OK:
+        log.warning("Post-update hook failed with exit code: %d", result.code)
+        result.code = EXIT.ERR_VM_POST_UPDATE
+    _log_output(log, result)
+    if print_streams:
+        _print_streams(result)
```
marmarek: `ProcessResult.process_communicate()` prints the output already, no? Or you may need to create `ProcessResult(realtime=True)` first, and then do something like `result += ProcessResult.process_communicate(proc)`. Anyway, my point is to not duplicate output printing.

marmarek (changes requested):

marmarek:
PipelineRetry

vmupdate/agent/source/common/post_update_hook.py
andreasglashauser: `ProcessResult.process_communicate()` doesnt already print it, so i took your second recommendation.

marmarek (approved):

